PR #1, opened 2016-06-14
Feature: Support fill page component (no dimensions)

vishakha94:
You can test this feature by commenting out dimensions property for xcomponent.create in checkout-components.
Logic??
1. Add a flag of 0 for both width and height if dimensions property not specified in checkout.js.
2. Check if options.dimensions == object? If not, give it (width, height) a number 0.
3. In dom.js changed the popup function: check if with,height==0. If yes, do window,open without passing any size for the window so as to have a new tab open.

### DIFF
--- a/src/component/component.js
+++ b/src/component/component.js
@@ -29,8 +29,18 @@ export let components = {};
 export class Component {
 
     constructor(options = {}) {
-        this.validate(options);
 
+        if (!options.dimensions || !(typeof options.dimensions === 'object')) {
+
+            options.dimensions = {
+                width : 0,
+                height : 0
+            };
+
+        }
+        
+        this.validate(options);
+        
         // The tag name of the component. Used by some drivers (e.g. angular) to turn the component into an html element,
         // e.g. <my-component>
 
--- a/src/component/component.js
+++ b/src/component/component.js
@@ -179,10 +179,11 @@ export class Component {
             throw new Error(`Invalid options.tag: ${options.tag}`);
         }
 
-        if (!options.dimensions || !(typeof options.dimensions === 'object')) {
+    /*    if (!options.dimensions || !(typeof options.dimensions === 'object')) {
             throw new Error(`[${options.tag}] Expected options.dimensions to be an object`);
         }
-
+    */
+        
         if (typeof options.dimensions.width !== 'number') {
             throw new Error(`[${options.tag}] Expected options.dimensions.width to be a number`);
         }
--- a/src/component/component.js
+++ b/src/component/component.js
@@ -30,17 +30,19 @@ export class Component {
 
     constructor(options = {}) {
 
-        if (!options.dimensions || !(typeof options.dimensions === 'object')) {
 
-            options.dimensions = {
-                width : 0,
-                height : 0
-            };
-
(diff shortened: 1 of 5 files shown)


src/component/component.js
```diff
@@ -166,10 +176,11 @@ export class Component {
             throw new Error(`Invalid options.tag: ${options.tag}`);
         }
 
-        if (!options.dimensions || !(typeof options.dimensions === 'object')) {
+    /*    if (!options.dimensions || !(typeof options.dimensions === 'object')) {
```
bluepnume: We can just remove this code, no need to leave it hanging around

src/lib/dom.js
```diff
@@ -39,6 +39,13 @@ export function getElement(id) {
 
 export function popup(url, options) {
 
+    // Condition for when dimensions (optional) for pop-up not specified
```
bluepnume: I have one small concern here -- if `width === 0` we're just ignoring all of the other options that might be passed too. I think we can probably do this without changing `popup()`

Probably if the code which calls `popup()` just does not pass in `width` and `height` this should work

Right now:

```popup({
  width: x,
  height: y,
  foo: 'bar'
})```

But if we change that to

```popup({
  foo: 'bar'
})```

We should get a full-screen popup

File: src/component/component.js [hunk 3]
```diff
@@ -29,8 +29,18 @@ export let components = {};
 export class Component {
 
     constructor(options = {}) {
-        this.validate(options);
 
+        if (!options.dimensions || !(typeof options.dimensions === 'object')) {
```
bluepnume: I think it might be better if we just leave `options.dimensions` blank here. Having a 0 width and height may be confusing for people reading the code in future.

bluepnume:
Added a few comments :) Can you also try to add a test for this case? If you run `gulp karma` you should be able to run the existing tests to make sure everything passes.

src/component/component.js
vishakha94: Cool!

src/lib/dom.js
vishakha94: Even better! Should i create a new commit?

src/component/component.js
vishakha94: Hmmmmm.

src/lib/dom.js
bluepnume: Maybe we can check if the width / height is a number *only* if they're passed?

```
    if (options.dimensions.width && typeof options.dimensions.width !== 'number') {
        throw new Error(`[${options.tag}] Expected options.dimensions.width to be a number`);
    }
```

vishakha94: ok..:)

vishakha94: since all the type checks are in the validate function it will be checked everytime. Probably I could take those two conditions out of validate and check on if options.dimensions is true. Else doing this:

if (options.dimensions.width && typeof options.dimensions.width !== 'number') {
        throw new Error(`[${options.tag}] Expected options.dimensions.width to be a number`);
    }

will trow errors

bluepnume: Yeah, we can do that

PR #28, opened 2016-06-24
Feature: AutoResizing of iframe

vishakha94:


src/component/child/index.js
```diff
@@ -31,6 +31,10 @@ export class ChildComponent extends BaseComponent {
         this.onProps = this.tryCatch(options.onProps || noop, false);
         this.onError = this.tryCatch(options.onError || (err => { throw err; }));
 
+        // watchforResize looks out for change in size of the contents inside of an iframe.
+        // It autoresizes the window if required.
+        this.watchForResize();
```
bluepnume: Can we do this in the `init()` method instead? Trying to keep instantiating and initializing the component to two different steps.

src/component/child/index.js
```diff
@@ -221,6 +225,34 @@ export class ChildComponent extends BaseComponent {
...
+    watchForResize() {
+        var elm = document.body;
+        var lastWidth = elm.scrollWidth, newWidth;
+        var lastHeight = elm.scrollHeight, newHeight;
+
+        setInterval(() => {
+            newWidth = elm.scrollWidth;
+            newHeight = elm.scrollHeight;
+            // Dimensions changed if this condition is true
+            if (lastHeight != newHeight || lastWidth != newWidth) {
```
bluepnume: Please use strict equality checks `!==`

src/component/child/index.js
```diff
@@ -221,6 +225,34 @@ export class ChildComponent extends BaseComponent {
...
+                var newDimensions = {
```
bluepnume: Can you please use `let` rather than `var`? Running `gulp lint` should show any other issues like this :)

bluepnume:
Looks good! Left a few comments, could you also add a unit test for this case?

Thanks!

bluepnume:
One more thing that came to mind -- can we make this configurable? Like I should be able to turn it on or off, as the component author:

https://github.com/krakenjs/xcomponent/issues/7

vishakha94:
Yup. We could do that. Cool idea.

bluepnume:
It would also be good to squash these commits into one so we end up with a clean commit for the feature:

http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit

src/component/child/index.js
```diff
@@ -235,6 +236,29 @@ export class ChildComponent extends BaseComponent {
...
+            if (lastHeight !== newHeight || lastWidth !== newWidth) {
+                let newDimensions = {
+                    width: newWidth,
+                    height: newHeight
+                };
+                postRobot.sendToParent(POST_MESSAGE.RESIZE, newDimensions);
+            }
+            lastWidth = newWidth;
+            lastHeight = newHeight;
+        }, 10);
```
bluepnume: Can we maybe make this 50ms? I'm concerned running this 100 times every second is going to be too much overhead.

src/component/child/index.js
```diff
@@ -235,6 +236,29 @@ export class ChildComponent extends BaseComponent {
...
+        setInterval(() => {
+            newWidth = elm.scrollWidth;
+            newHeight = elm.scrollHeight;
+            // Dimensions changed if this condition is true
+            if (lastHeight !== newHeight || lastWidth !== newWidth) {
+                let newDimensions = {
+                    width: newWidth,
+                    height: newHeight
+                };
+                postRobot.sendToParent(POST_MESSAGE.RESIZE, newDimensions);
```
bluepnume: Can we just use `this.resize(width, height)` here?

src/component/parent/drivers.js
```diff
@@ -150,6 +150,7 @@ export let RENDER_DRIVERS = {
 
             this.iframe.style.zIndex   = MAX_Z_INDEX;
             this.iframe.style.position = 'fixed';
+            this.iframe.className  = 'xcomponent-lightbox bounceInUp';
```
bluepnume: Can we avoid hard-coding `bounceInUp` here? I'm not sure every component is going to want this behavior. It might be best just to keep the `xcomponent-lightbox` class and have the component add the bounce CSS under that class.

src/component/parent/index.js
```diff
@@ -561,7 +561,11 @@ export class ParentComponent extends BaseComponent {
 
             [ POST_MESSAGE.CLOSE ](source, data) {
 
-                this.close();
+                this.iframe.className = 'xcomponent-lightbox bounceOutDown';
+
+                setTimeout((() => {
```
bluepnume: We need to think carefully about this one. `setTimeout` should really be avoided since in a lot of cases (especially in IE) if the window isn't focused, it will hang forever.

We also probably don't want to hard code a 1000ms delay here, since this bounce-out behavior is pretty specific to only one of our components.

I don't know what the ideal solution is here, but I'll have a think about it. In the mean time maybe we can just remove this timeout.

src/component/parent/index.js
```diff
@@ -561,7 +561,11 @@ export class ParentComponent extends BaseComponent {
 
             [ POST_MESSAGE.CLOSE ](source, data) {
 
-                this.close();
+                this.iframe.className = 'xcomponent-lightbox bounceOutDown';
```
bluepnume: Few points here:

1. Again would be a little hesitant to hard code `bounceOutDown` here. Maybe we can just have an `xcomponent-closing` class so users can define whatever animation they want under that?

2. `this.iframe` will not always be available. I'd recomment putting any custom `close` code in `drivers.js` then doing something like

```
if (RENDER_DRIVERS[this.context].close) {
    RENDER_DRIVERS[this.context].close.call(this);
}
```

I'd also recommend calling running this code ^ from the `close()` method rather than here.

src/component/parent/index.js
```diff
@@ -614,6 +619,17 @@ export class ParentComponent extends BaseComponent {
...
     }
 
+    /*  Reposition
+     ----------
+
+       Reposition the child component window after Auto Resize
+     */
+
+    reposition() {
+        this.iframe.style.top = window.innerHeight / 2 - this.iframe.height / 2;
```
bluepnume: Maybe we could just take all of this code:

https://github.com/vishakha94/xcomponent/blob/master/src/component/parent/drivers.js#L155-L175

And put it in a 'reposition' method, then call it from both places. Also it would be good to keep this in `drivers.js` since it's pretty specific to the iframe.

In general, as a rule of thumb, if there's any code where we're doing `if (this.iframe)` or `if (this.context === x)` it's best to put it in `drivers.js`, anything that's specific to iframes or popups or lightboxes should live there so parent.js doesn't have to deal with too much branching logic.

src/component/parent/index.js
```diff
@@ -625,6 +641,7 @@ export class ParentComponent extends BaseComponent {
         if (this.iframe) {
             this.iframe.height = height;
             this.iframe.width = width;
+            this.reposition();
```
bluepnume: Would recommend making this a driver method, like:

```if (RENDER_DRIVERS[this.context].reposition) {
    RENDER_DRIVERS[this.context].reposition.call(this);
}```

bluepnume:
Hi Vishaka, added a few more comments -- mostly just code organization, the logic looks really good :) thanks for taking care of this!

Few more things that didn't fit into any other specific place:

1. For features like this, can you please squash commits? http://eli.thegreenplace.net/2014/02/19/squashing-github-pull-requests-into-a-single-commit Then the history is clean, and we keep tests passing from one commit to another

2. Can you add the `autoResize` option (See https://github.com/krakenjs/xcomponent/issues/7) before we merge this in? This should probably default to `false`, since many components are likely to just be a static size.

Thanks!
Daniel

src/component/child/index.js
vishakha94: Done.

src/component/parent/drivers.js
vishakha94: So I have renamed the class to xcomponent-lightbox-open. Using just xcomponent-lightbox and naming the animation: bounceInUp in it was creating some issues(overlay wasn't created and the box got cut off). Nonetheless I hope it resolves the bounceInUp hard coding issue.

bluepnume: Sounds good!

bluepnume:
Looks perfect! Merging :)

vigosan:
@bluepnume Is this feature still working on master? I'm trying to embed using an iframe, but it is not auto resized:

screen shot 2017-06-14 at 2 47 29 pm

screen shot 2017-06-14 at 2 49 13 pm

Am I missing something in the configuration? thx

PR #55, opened 2017-03-07
feat(sourcemap): Add Source maps

trainerbill:
For troubleshooting


bluepnume:
Looks like there's a conflict. Could you remove any changes to dist/ then they'll be picked up when the code is published. Makes it easier to track diffs etc. if we keep dist/ changes to their own isolated commits.

trainerbill:
done

PR #73, opened 2017-03-30
Update index.js to allow resizing when not using autoresize

mmairs9:
The auto resize does not always work the best especially when shrinking/growing width of the browser. We would however like to tell the iframe to resize itself in certain scenarios without having to have this autoResize functionally.


bluepnume:
Looks good to me!

Issue #74, opened 2017-03-31
Question: how do I get rid of all the logs created by xcomponent in the console? Is there a prop I can set?


mmairs9:

bluepnume:
Hey @mmairs9 -- most of the functionality for this is already built as part of https://github.com/paypal/paypal-checkout/issues/180, I just need to expose it as an option when creating xcomponents. I'll keep you posted.

raphaelcosta:
👍

harouny:
@bluepnume 
Hi Daniel, 
I started to work on this item. I had a question: Is the intention is to use props as the only way of communication between host and components? so for example will we have (logLevel) as an internal prop with the default 'info' then when then host wants to change that it pass ('warn', 'error', 'debug') as a value for logLevel prop?

bluepnume:
Hey @harouny -- that's exactly right. If you're looking for an example, here's how I'm doing it over in `paypal-checkout`: https://github.com/paypal/paypal-checkout/commit/a675143f40e77aa52614e83b0d33d43e00db9229

I'm thinking log-level should be an option when you first create the component, and that should be the default log-level that's passed down to the child.

harouny:
@bluepnume cool, I will submit a PR soon

bluepnume:
Fixed and published, thanks for the fix @harouny!

PR #80, opened 2017-04-24
Ability to configure console log level. Issue #74

harouny:


bluepnume:
Hey this is awesome. Couldn't find a single problem with your code, thanks for making the contribution!

Just one quick request before I merge -- could you back out the changes to `dist/`? I prefer to keep changes to `dist/` separate from changes to `src/` so I can keep distributed code in sync with published npm versions, and also keep the git history clean so it's easy to revert changes in `src/` if needed. Once I've merged, I'll build and publish.

Thanks again!

harouny:
@bluepnume no worries mate. I removed dist changes.

bluepnume:
Coolcool, merged and published!

PR #84, opened 2017-05-04
Fix for validation error breaking Remote demo

thirtified:
Added missing 'px' unit to dimensions specification.


bluepnume:
Looks good, thanks!

PR #90, opened 2017-05-19
PR for allowing option to opt in to not immediantly destroy the xcomponent when parent navigates to a new page

mmairs9:


bluepnume:
Ended up with a different fix, per https://github.com/krakenjs/xcomponent/issues/89

mmairs9:
Thanks, feels a lot better with this change in.

Issue #52, opened 2016-12-21
Add domain check before sending init message to parent


bluepnume:

harouny:
@bluepnume Can you add more info to that item please? What kind of domain check still missing?

bluepnume:
Hey, the intent of this was to allow components to define which domains they can be rendered by. For example, if a component specified:

```
parentDomains: [
    'https://a.com',
    'https://b.com'
]
```

but the component was rendered by `https://c.com`, the component should error out and not send any init message (or any message) up to the parent.

harouny:
Cool. Thanks for the info. On that just to confirm, if a component didn't specify parentDomains, it means that it can send messages to any domain.

bluepnume:
Yeah, the way it stands right now is that components can be rendered by any domain. If you need to restrict this the best way right now is to rely on an `X-FRAME-OPTIONS` header.

PR #93, opened 2017-05-26
Feature/add domain check before sending init message to parent #52

harouny:
add domain check before sending init message to parent #52


harouny:
@bluepnume Hey, this change should prevent child from sending any msg to parent. Its currently not very graceful experience since child was already rendered.  But that's what I understood from #52 anyway, that we need to prevent sending init to parent. I hope my understanding was correct.

harouny:
Actually I'm doing a small change here. If the item in the array is a string, its better to do a full string match instead of excpecting the string to be a RegEx string...this will give people a simpler string interface while RegEx still supported by passing Regex items.

harouny:
@bluepnume Hey mate,
Do you know when you plan to push this in? I'm happy to change anything here :)

bluepnume:
Hey @harouny, just out of the country on vacation right now. Back on the weekend -- I'll give this a proper review then. Thanks so much for the contribution!

harouny:
@bluepnume
Sorry, didn't know that since you are already contributing to discussions.
Enjoy your vacation and don't check your email :)

src/component/child/index.js
```diff
@@ -76,6 +77,28 @@ export class ChildComponent extends BaseComponent {
         });
     }
 
+    validateParentCommunication() {
+        let isAllowed = false;
+        if (this.component.allowedParentDomains && this.component.allowedParentDomains.length && this.component.allowedParentDomains.length > 0) {
+            const parentDomain = this.getParentDomain();
+            each(this.component.allowedParentDomains, (parentDomainToMatch) => {
```
bluepnume: Can you use `matchDomain` from `cross-domain-utils` for this?

```
if (matchDomain(this.component.allowedParentDomains, this.getParentDomain())) {
   ...
}
```

src/component/child/index.js
```diff
@@ -53,6 +53,7 @@ export class ChildComponent extends BaseComponent {
         //
         // - What context are we
         // - What props has the parent specified
+        this.validateParentCommunication();
```
bluepnume: Let's run this check right up on line 30 before anything has a chance to initialize. And let's error the component out.

```
if (!this.hasValidParentDomain()) {
    return this.error(new Error(`Can not be rendered by domain: ${this.getParentDomain()}`));
}
```

src/component/component/index.js
```diff
@@ -41,6 +41,8 @@ export class Component extends BaseComponent {
 
         this.addProp(options, 'defaultLogLevel', 'info');
 
+        this.addProp(options, 'allowedParentDomains', []);
```
bluepnume: Let's default this to `'*'` so `checkDomain` matches everything by default.

docs/api.md
```diff
@@ -70,6 +70,17 @@ dimensions: {
 }
 ```
 
+#### allowedParentDomains `Array<string | RegEx>`
+
+An array of strings or reqular expresions to be used to validate parent domain against. If provided, and parent domain doesn't match any item, communication from child to parent will be prevented.
```
bluepnume: String, array of strings, or regular expression.

bluepnume:
Added a few comments. Could you also add a test or two for this case?

harouny:
@bluepnume I want to add couple of happy path tests, do you know what will ```getParentDomain()``` return in test?

bluepnume:
You can set `window.mockDomain` to `mock://something` and xcomponent/post-robot will all pick it up

harouny:
Hey @bluepnume 

Looks like ```window.mockDomain``` is not enough in my case. I was getting this error:
```
ERROR LOG: 'xc_test_component_parent_domains_string_match_unexpected_listener_init', Object{origin: 'http://localhost:9876', domain: 'mock://foo.com', timestamp: 1497050613115, windowID: '7d58e48fae', pageID: '5407a6bcce'}
```
I think the problem in origin check?

I ended up for now relying my test on running on ```http://localhost:9876```  based on karma config. Will be better to avoid that.

src/component/child/index.js
```diff
@@ -76,6 +80,23 @@ export class ChildComponent extends BaseComponent {
         });
     }
 
+    hasValidParentDomain() {
+        const parentDomain = this.getParentDomain();
+        if (typeof this.component.allowedParentDomains === 'string') {
```
bluepnume: You can use `matchDomain()` for arrays and regexes too, so no need to code a loop here at all

harouny: @bluepnume Ok, will check that out

harouny: Given results of these tests, ```matchDomain ``` needs some attention:
``` javascript
import { matchDomain } from 'cross-domain-utils/src';

describe('match domain', () => {
    describe('returns true when', () => {
        it('domain is array of regex and origin is array of regex', () => {
            assert.isTrue(matchDomain([/^http\:\/\/domain$/], [/^http\:\/\/domain$/]));
        });

        it('domain is a string and origin is a string', () => {
            assert.isTrue(matchDomain(['http://domain'], 'http://domain'));
        });

        it('domain is array of regex and origin is a string', () => {
            assert.isTrue(matchDomain([/^http\:\/\/domain$/], 'http://domain'));
        });

        it('domain is regex and origin is a string', () => {
            assert.isTrue(matchDomain(/^http\:\/\/domain$/, 'http://domain'));
        });
        
        it('domain is array with a wildcard and origin is a string', () => {
            assert.isTrue(matchDomain(['*'], 'http://domain'));
        });

        it('domain is array of strings and origin is a string', () => {
            assert.isTrue(matchDomain(['http://domain'], 'http://domain'));
        });
        
    });
});
```
match domain
    returns true when
      √domain is array of regex and origin is array of regex
      √domain is a string and origin is a string
      ×domain is array of regex and origin is a string
        expected false to be true
      ×domain is regex and origin is a string
        expected [ Array(1) ] to be true
      ×domain is array with a wildcard and origin is a string
        expected false to be true
      √domain is array of strings and origin is a string
```

```

bluepnume: Thanks for the matchDomain tests! Added and fixed in cross-domain-utils, can you try with the latest version?

bluepnume:
Also just had another thought -- we should probably also run this check during render() on the parent page, so we fail as early as possible.

harouny:
I agree. Already mentioned this in my first comment in this PR :)

src/component/child/index.js
harouny: Works fine after your fix. Now I pass the array to matchDomain.

harouny:
I'm currently working on validating that in parent render function too.

bluepnume:
Looks good to me! Ready for merge?

harouny:
I think so...fingers crossed :)

bluepnume:
Awesome! Thanks again for taking this one up!

harouny:
No worries mate...thanks for your help.

PR #98, opened 2017-06-14
fix angular.js minification

david-gang:
Hi,

Currently the minified js file does not work minified with angular.
https://github.com/paypal/paypal-checkout/issues/371

This is also connected: https://github.com/paypal/paypal-checkout/issues/288

The fix is quite straight forward.

Thanks,
David


bluepnume:
Thanks for the fix!

david-gang:
It was a pleasure :-)

Issue #23, opened 2016-06-18
Support for more frameworks


bluepnume:
Need to complete building drivers for:
- [x] Javascript
- [x] Script tag
- [x] Angular
- [x] React
- [x] Glimmer
- [x] Vue
- [x] Angular 2.0
- [ ] Ember
- [ ] Polymer
- [ ] Backbone
- [ ] Custom HTML element?

Others?

davidmathlogic:
I'd certainly like to see support for Vue.js.

bluepnume:
Sounds like a plan 👍

adilmughal:
Angular 2+ support please

harouny:
@bluepnume I started a POC for Angular 2.0 support. I know its not as straightforward as the others.
First item done is to have a minimal setup for an angular 2 app in demo/frameworks/angular2 that we can use in dev 
https://github.com/harouny/xcomponent/tree/feature/angular2_driver/demo/frameworks/angular2

harouny:
One goal is to not mess up with our package.json etc with Angular 2.0 / Typescript stuff. So everything is compiled/transpiled in browser.

veenasubramanyam:
Is there an example or demo of how to embed a react component inside an angular 2 container?

bluepnume:
Angular 2.0 isn't yet supported. But you could just do something like this:

https://github.com/krakenjs/xcomponent/tree/master/demo/advanced/react-end-to-end

In `index.htm`, instead of using React, you could just do `MyLoginXComponent.render({ ... }, '#container')` in an angular 2 component.

harouny:
Hey @bluepnume 

I have done good progress in angular 2 driver this weekend. However something worth a discussion:

Angular team decided at some point to have the following naming:
AngularJs is the name of versions < 2
Angular is the name of verstions >= 2

In Xcomponent we have '''angular''' currently reserved for AngularJs versions < 2 (currently in version 4 afaik).

We have the following options for the name of Angular > 2:

- angular (requires a breaking change to rename existing angular to angularJs)
- angular2 (will indicate angular 2+)
- ng (angular team used ng in places where angular was reserved i.e window.angular)

I feel like angular2 is not too bad, but needed other opinions.

morbidick:
Hi @harouny and @bluepnume,
is there any documentation on the api the driver has to provide? For `register` its more or less clear what has to happen, but what about `scan` since thats probably the hard part to get shadow dom working with polymer?
Thx in advance!

[...]

PR #102, opened 2017-06-21
Feature/angular2 driver (#23)

harouny:
Hey @bluepnume

I expect a bit of discussion around this feature, no need to rush here. Feel free to send through any comments.

Things worth to discuss:
The use of ```driverResults```.
Having to do ```bind(this)``` in host angular2 component.

My next step here is to work on loose props support, and add a few tests.


bluepnume:
This is really awesome -- thanks for the contribution! 

I'll try to take some time before the end of this week to do a thorough review.

harouny:
@bluepnume Thanks, I got loose props working and also added canBeRenderedByDriver check to prevent automatically registering all drivers if components don't support rendering into an Iframe, please confirm my assumption.

src/drivers/angular2.js
```diff
@@ -0,0 +1,99 @@
+import { omit, replaceObject } from '../lib';
+/* eslint-disable new-cap, object-shorthand */
+
+
+export let angular2 = {
+
+    global() {
```
bluepnume: Could we just return `window.ng` at this point? That also means someone programmatically generating the driver could just call:

```javascript
MyComponent.driver('angular2', ng);
```

Rather than needing to pass in all of the below child dependencies.

src/component/component/index.js
```diff
@@ -137,19 +137,31 @@ export class Component extends BaseComponent {
...
+
     registerDrivers() {
         this.driverCache = {};
 
         for (let driverName of Object.keys(drivers)) {
+            this.driverResults[driverName] = {};
             let driver = drivers[driverName];
             let glob = driver.global();
-            if (glob) {
+            if (glob && this.canBeRenderedByDriver(driverName)) {
```
bluepnume: Let's just make this `this.contexts[CONTEXT_TYPES.IFRAME]` rather than going down the try/catch route.

src/drivers/angular2.js
```diff
@@ -0,0 +1,99 @@
...
+            constructor: function () {
+                
+            }
+        });
+        
+        xcomponent.driverResults.angular2 = {
+            module: Angular2Module
+        };
+
+        return xcomponent;
```
bluepnume: Can we return `Angular2Module` at this point? Then people can reference it with either: `MyComponent.angular2` or `MyComponent.driver('angular2', ng)`

src/component/component/index.js
```diff
@@ -137,19 +137,31 @@ export class Component extends BaseComponent {
         // Register all of the drivers for instantiating components. The model used is -- there's a standard javascript
         // way of rendering a component, then each other technology (e.g. react) needs to hook into that interface.
         // This makes us a little more pluggable and loosely coupled.
-
+        this.driverResults = {};
```
bluepnume: I think if we return `Angular2Module` in the driver, we can avoid exposing `driverResults`, right?

src/drivers/angular2.js
bluepnume: OK I just tried this, and it seemed to work pretty well:

```javascript
      var HelloComponent = ng.core.Component({
          selector: 'hello-world',
          template: `
              <b>hello world!</b>
          `
      }).Class({
          constructor: function() {}
      });

      var AppComponent = ng.core.Component({
          selector: 'my-app',
          template: `
              <h1>My App</h1>
              <hello-world></hello-world>
          `
      }).Class({
          constructor: function() {}
      });

      var AppModule = ng.core.NgModule({
          imports: [ ng.platformBrowser.BrowserModule ],
          declarations: [ AppComponent, HelloComponent ],
          bootstrap: [ AppComponent ]
      }).Class({
          constructor: function () { }
      });
```

Notice how `HelloComponent` is being imported directly into the `AppModule`. This means we can keep the xcomponent driver a bit simpler and just export the component, rather than having to create a new angular module and export that. Work for you?

src/drivers/angular2.js
```diff
@@ -0,0 +1,99 @@
...
+                for (let key of Object.keys(xcomponent.props)) {
+                    inputs.push(key);
+                }
+            }
+            return { inputs };
+        };
+
+
+        function getProps(component) {
+            const props = component.props || omit(component, ['$xContext']);
```
bluepnume: Not sure I like the `omit(component, ['$xContext'])` here. I'd prefer to map over the list of known prop names and extract them from `this`.

src/drivers/angular2.js
```diff
@@ -0,0 +1,99 @@
...
+            BrowserModule: window.ng.platformBrowser.BrowserModule
+        } 
+        : false;
+    },
+
+    register(xcomponent, { Component, NgModule, ElementRef, NgZone, BrowserModule }) {
+
+        xcomponent.log('initializing angular2 component');
+        
+        const getBindingMetadata = () => {
```
bluepnume:
```javascript
return xcomponent.looseProps ? [ 'props' ] : Object.keys(xcomponent.props);
```

src/drivers/angular2.js
```diff
@@ -0,0 +1,99 @@
...
+            .Class({
+                constructor: [ElementRef, NgZone, function(elementRef, zone) { 
+                    this.$xContext = {
+                        elementRef,
+                        zone
+                    };
+                }],
+                ngOnInit: function () {
+                    const targetElement = this.$xContext.elementRef.nativeElement;
+                    const parent = xcomponent.init(getProps(this), null, targetElement);
```
bluepnume: The fact that angular still doesn't have a dedicated `props` property or something to contain props makes me sad. All of this is making me realize just how much I love React...

src/drivers/angular2.js
```diff
@@ -0,0 +1,99 @@
...
+                    parent.render(targetElement);
+                    this.$xContext.parent = parent;
+                },
+                ngOnChanges: function() {
+                    if (this.$xContext.parent) {
```
bluepnume: Would there ever be a case when we don't expect to have `this.$xContext.parent`?

src/drivers/angular2.js
```diff
@@ -0,0 +1,99 @@
...
+            })
+            .Class({
+                constructor: [ElementRef, NgZone, function(elementRef, zone) { 
+                    this.$xContext = {
+                        elementRef,
+                        zone
+                    };
+                }],
+                ngOnInit: function () {
+                    const targetElement = this.$xContext.elementRef.nativeElement;
```
bluepnume: Is it not possible to inject `ElementRef` here too, rather than needing to persist it in `$xContext` ?

bluepnume: Hmm, from a little reading I guess the recommended pattern is just to do all of this in the constructor.

In that case, can we use the `__XCOMPONENT__` constant instead of `$xContent`? So

```javascript
this[__XCOMPONENT__] = { element, zone };
```

demo/frameworks/angular2_ts/app-component.html
```diff
@@ -0,0 +1,7 @@
+<div>
```
bluepnume: Can we call this directory `angular2-typescript` for clarity's sake?

demo/frameworks/angular2/index.htm
```diff
@@ -0,0 +1,77 @@
...
+    <script src="../../../dist/xcomponent.js"></script>
+    <script src="./login.js"></script>
+</head>
+
+<body>
+    <my-app>Loading...</my-app>
+    <script>
+
+        // component
+        (function(app) {
```
bluepnume: Let's just put everything in one big IIFE, and use local variables rather than an `app` global, just to make the example as simple as possible for people to read through.

demo/frameworks/angular2/index.htm
```diff
@@ -0,0 +1,77 @@
...
+            app.AppComponent =
+                ng.core.Component({
+                selector: 'my-app',
+                template: `
+                        <div>
+                            <h3>Log in on xyz.com</h3>
+                            <div id="container">
+                                <my-login-component [props]="props"></my-login-component>
```
bluepnume: Debating whether we should show the expanded props in the demo -- like [onLogin]="onLogin". 

Sucks that we need to have two different ways of passing props in the first place. I wish angular allowed variable props like react's `<FooComponent {...props} />` which would make this a moot point.

bluepnume: Actually, here's what I'm thinking, so we can just have "one way of doing things™":

1. Let's disallow patterns like `[onLogin]="onLogin"` and only allow `[props]="props"`
2. Then in the demo, the example we show can be like: `[props]="{ prefilledEmail: email, onLogin: onLogin }"`
3. We can do the same for angular1

That way:
- We get a consistent `props` object rather than needing to loop through `this` to extract the props.
- You can still see in the template which props are being passed down
- There aren't multiple ways of rendering a component with props depending on if the component has pre-registered props or loose props.

I think that's an easy win.

bluepnume:
OK, left a few comments but I'm generally very happy with the direction this is taking. Thanks again for putting in all the hard work, and sorry for the delay reviewing -- had to learn some angular 2 first! :)

src/drivers/angular2.js
bluepnume: Or on second thoughts -- if we go with the "Only register `props` in `inputs`" approach that I was talking about above, we can avoid having this context object entirely, and set these dependencies on `this` directly, since then there would be no chance of those names conflicting with prop names.

demo/frameworks/angular2/index.htm
harouny: I thought we needed to support both method since both methods already supported in other drivers.

To be honest communicating between components through one (props) object is only natural to react people. Angular people (99% cases using typescript) would expose their props as public properties to component class and won't creat another object/type to hold props.


Related to that in angular functions are not passed down -technically possible but considered a bad practice and unnatural-. Instead event emitters emit events up and parent component handles it using an event binding: (OnLogin)="Login($event)"  Implementing that is in xcomponent is also possible and even having the two techniques configurable. 

I guess what I mean is providing natural and excpected model to people using a specific framework is more important that implementing all frameworks the same way.

src/component/component/index.js
harouny: Its possible, actully I would have prefered that, however, it can be considered a breaking change scince same function returns the component itself in other drivers. i.e. angularJS

harouny: Actually in umd mode where globals are ready and automatic registeration happend. Users do call this function directly so we still need to save the module somewhere. Best location is the component itself scince its already global.

harouny: sorry "Users do *not call this function"
typing on mobile :)

src/drivers/angular2.js
harouny: There is no equivlant to window.ng in non umd mode.

harouny: non umd mode will be 99% of the cases. Check typeScript example.

src/drivers/angular2.js
harouny: The only reason I kept it like that is that I was thinking of declaring outputs (event emitters) in a later feature based on type of the prop is function. But maybe we can worry about that later.

src/drivers/angular2.js
harouny: We dont have a list of known props in loose props mode. Unless if we only communicate via a single props object or dumb loose props support list of known props is not applicable here.

src/component/component/index.js
bluepnume: >  it can be considered a breaking change scince same function returns the component itself in other drivers

Yeah looks like the original `angular` component is doing that, but we can remove that with the next major version bump.

We won't be breaking any backwards compatibility for angular2 since nobody is using it yet, so I think we're safe to return the `Angular2Component` directly.

demo/frameworks/angular2/index.htm
bluepnume: Only problem is we need to support the loose-props case too. So unless we have a single `props` object, we end up having to support two different ways of passing down props for the same driver. I'd prefer to normalize that down to one way.

src/drivers/angular2.js
harouny: Yes, the first time binding happens...before ngOnInit gets called.

src/drivers/angular2.js
bluepnume: Ah yep, you're totally right about loose props...

src/drivers/angular2.js
bluepnume: Fair point. Wanna mull on this some more though... feels a little like the moment we need any additional small dependency in future, we'd have to make a backwards incompatible change to this driver.

src/drivers/angular2.js
bluepnume: OK

demo/frameworks/angular2/index.htm
bluepnume: I haven't played around with event-emitters yet (more abstractions over vanilla functions seems like an anti-pattern to me) but for sure, let's see how well this could fit in especially with loose props.

harouny: To be honest my implemention of loose props here doesn't relay on having a single props object. You can simply pass no props to xc.create and still pass props individually through proprty binding. It will work perfectly, the only thing left is user need to pass a flag to angular to skip erroring when it sees an unknown property.

src/drivers/angular2.js
harouny: Yes very annoying.

Import * as core from '@angular/core'
That would solve big part of the problem however will be annoying to angular users used to import core functions individually. Your thoughts?

bluepnume: Potentially yeah. Especially if we can export a component directly and avoid requiring `BrowserModule`.

demo/frameworks/angular2/index.htm
bluepnume: Hmm, interesting, didn't know about that. `schemas: [ ng.core.NO_ERRORS_SCHEMA ]`. Shame it only works when passed to a module, not for an individual component...

harouny: Yep, and that must be the module that is rendering the component (in the user side) not the module that is exporting it.

src/drivers/angular2.js
harouny: We just need to check how that impacts angular's ahead of time compilation. One more thing for you to not like in angular :)

src/drivers/angular2.js
harouny: First I don't think we have to import ```ng.platformBrowser.BrowserModule``` in our wrapping module, I will check that out. 
Second, I still feel like we should be passing a module instead of a component. As in future we may need to add a service or a pipe .etc and only passing a component won't help. Users in angular world used to share plugins and components via modules due to the same reason. As an example: https://material.angular.io/guide/getting-started where each UI component is imported via importing its module.

bluepnume: Sure, I'm open to the idea of returning a module -- you're right, there are definitely some future-proofing benefits there and it does seem more standard in the angular world.

bluepnume: Plus it's more of a natural progression from the angular 1 driver, which by necessity has to create a module.

demo/frameworks/angular2/index.htm
harouny: OK, will do.

demo/frameworks/angular2_ts/app-component.html
harouny: I will do that.

src/component/component/index.js
harouny: Then just confirming that with the approach of returning the module from register method plus setting ```component.driverResults.angular2``` or ```component.angular2``` to get an accessor in automatic mode  approach we have two ways of accessing driver results. While if we stick to always setting ```component.driverResults.angular2``` or ```component.angular2``` its one way that work both cases.

src/component/component/index.js
harouny: Yep, I can do that. Or to keep context validation in one place we can have two methods ```validateRenderContext``` that throws the error after calling ```canBeRenderedByContext``` that only return true or false.

src/component/component/index.js
bluepnume: Not sure I follow -- how is it one way?

This is what we have right now for :

Automatic: `let frameworkComponent = MyComponent.framework;`
Manual: `let frameworkComponent = MyComponent.driver('framework', framework);`

If we add in `driverResults` don't we end up with 3 different ways?

Automatic: `let frameworkComponent = MyComponent.framework;`
Automatic 2: `let frameworkComponent = MyComponent.driverResults.framework;`
Manual: `let frameworkComponent = MyComponent.driver('framework', framework);`

demo/frameworks/angular2/index.htm
bluepnume: Yeah. So while I guess that works, I don't think it's a viable option for us to recommend here, as it loses people a bunch of pretty useful error checking for all of their other components.

src/component/component/index.js
harouny: I meant that:
```
imports: [MyLoginXComponent
  .driverResults
  .angular2
  .module]
```

Is accessing the module the same way as:
```
imports: [MyLoginXComponent
   .driver('angular2', { Component, NgModule, ElementRef, NgZone, BrowserModule})
  .driverResults
  .angular2
  .module]
```

and that:
```
MyLoginXComponent
   .driver('angular2', { Component, NgModule, ElementRef, NgZone, BrowserModule});

imports: [MyLoginXComponent
  .driverResults
  .angular2
  .module]
```
The only difference is not initializing the driver because it has been done already. 

Looks like I haven't seen the implementation of all drivers and looks like they are different because I haven't seen this in angular driver:
 ```
let frameworkComponent = MyComponent.frameworkName;
```

Comment (bluepnume) [inline src/component/component/index.js]: In your second example:

```
imports: [MyLoginXComponent
   .driver('angular2', { Component, NgModule, ElementRef, NgZone, BrowserModule})
  .driverResults
  .angular2
  .module]
```

Providing we return the autogenerated angular2 module in `register()`, this should be:

```
imports: [
    MyLoginXComponent.driver('angular2', {
        Component, NgModule, ElementRef, NgZone, BrowserModule
    })
]
```

The fact that the angular 1 driver returns the xcomponent in `register()` is a mistake -- it doesn't add any value. It should probably just return undefined. But for angular 2 the most useful thing for us to return is the module we create.

Comment (harouny) [inline src/component/component/index.js]: OK cool, I got you now.

Comment (harouny) [inline src/component/component/index.js]: Should we then change: ```registerDrivers``` method to have:
```
if (glob) 
{
       this[driverName] = this.driver(driverName, glob);
}
```
instead of:
```
if (glob) 
{
       this.driver(driverName, glob);
}
```

File: src/component/component/index.js (hunk 2 above)
Comment (bluepnume) [inline src/component/component/index.js]: Cool. Either works for me 👍

File: src/component/component/index.js (hunk 4 above)
Comment (bluepnume) [inline src/component/component/index.js]: If we're removing that line we might as well remove the entire `registerDrivers` method to avoid auto-registering any of these drivers. That's the direction I think I'd like to head -- but if it's ok with you, let's leave it for another PR since it's a backwards-incompatible change.

**Comment by harouny** [comment]
@bluepnume No worries I will go through changes we discussed today, most likely upcoming weekend. This was a way for me also to know more about using angular 2 without TypeScript :)

File: src/drivers/angular2.js (hunk 1 above)
Comment (harouny) [inline src/drivers/angular2.js]: Resolved by:
```javascript
import * as ngCore from '@angular/core';
const MyLoginXComponentModule = MyLoginXComponent.driver('angular2', ngCore);
```

src/component/component/index.js
harouny: Resolved by removing this check completely since its now out of scope of this change.

src/drivers/angular2.js
harouny: Resolved by not relying on ```ng.platformBrowser.BrowserModule```
And exporting angular module

src/component/component/index.js
harouny: Resolved by removing ```driverResults```, returning angular module and removing automatic driver registration.

src/drivers/angular2.js
harouny: Resolved by only using ```props``` in binding.
```
<my-login-component [props]="{ onLogin: onLogin, prefilledEmail: prefilledEmail }"></my-login-component>
```

src/drivers/angular2.js
harouny: Resolved by removing ```getBindingMetadata``` and only using ```props``` in binding

src/drivers/angular2.js
harouny: Resolved by only using ```props``` in binding.

src/drivers/angular2.js
harouny: Resolved by only using ```props``` in binding.

demo/frameworks/angular2_ts/app-component.html
harouny: Resolved

demo/frameworks/angular2/index.htm
harouny: Resolved

demo/frameworks/angular2/index.htm
harouny: Resolved by only using ```props``` in binding.

harouny:
@bluepnume 
I have addressed issues as discussed, let me know if you have any comments.

src/drivers/angular2.js
```diff
@@ -0,0 +1,68 @@
...
+    register(xcomponent, { Component, NgModule, ElementRef, NgZone }) {
+
+        xcomponent.log('initializing angular2 component');
+        
+        function getProps(component) {
+            return replaceObject(component.props, (value, key, fullKey) => {
+                if (typeof value === 'function') {
+                    return function () {
+                        let result;
+                        component.zone.run(() => {
```
bluepnume: Just a minor point but I think you can just return this directly:

```javascript
return component.zone.run(() => value.apply(this, arguments));
```

bluepnume:
This is looking really good -- I'm totally happy with it. Awesome work! Couple of tests and I think it'll be totally ready to merge, unless there's anything else you're waiting to add.

src/drivers/angular2.js
bluepnume: Cool, looks good!

src/drivers/angular2.js
harouny: Nice! I didn't know it returns the value returned from the callback

harouny:
@bluepnume Thanks mate, I'm already working on tests now. I don't think we need to add more to this one at this stage.

bluepnume:
Tests look good! I'm pretty much ready to merge this if you're good to go.

Couple of ideas for later:

- Would be good to have one or two integration tests which actually render the Angular 2 component and make sure props are called correctly, etc.

- Thoughts on writing a blog post announcing this? Could talk about some examples and some of the challenges / quirks.

harouny:
I couldn't yet find a nice angular2 way to do this:
$compile('my html string')
and that blocked my integration test attempt

The blog post is a good idea, I would like to do that maybe in the next few days. Will post on my meduim feed.

There is two unit tests in my mind for ngOnChanges and for zone.run but I can add those later...they won't require publishing. So I'm good to merge now.

harouny:
Hey @bluepnume 
A URL for a draft version of the blog post:
https://medium.com/@harouny/introducing-support-for-angular-2-in-xcomponent-cb31a89f01
Have a look and send through any comments

bluepnume:
Looks great to me!

Could you just link the first mention of `xcomponent` to this repo?

harouny:
@bluepnume 
Ok I will. 
I noticed dist and package.json haven't been updated here yet. Just wanted to make sure you published the npm package before I publish that post.

bluepnume:
Sorry for the delay -- had a few breaking changes I wanted to get in before a major version bump. Published v3 with your driver!

harouny:
No worries. I will update my post to mention version number and publish.

harouny:
On that, what's the list of breaking changes? It should be published somewhere.

bluepnume:
Yeah, it really should. I'm gonna try to find some time this week to throw up a changelog, it's high time... https://github.com/krakenjs/xcomponent/issues/108

PR #119, opened 2017-09-27
Add autoResize flag to the `create` api docs

namirsab:
As the title says, adding just a small bit of documentaton of a really helpful feature.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -70,6 +70,17 @@ dimensions: {
 }
 ```
 
+#### autoResize `boolean`
+
+When set to `true`, makes the xcomponent parent iframe resize automatically when the child component size changes
+
+```javascript
+dimensions: {
+    width: '300px',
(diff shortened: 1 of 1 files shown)


bluepnume:
Looks great -- could you also add a note about enabling for just height/width? For example, it's possible to do:

```javascript
autoResize: {
    width: false,
    height: true
}
```

Thanks again!

twig-openlearning:
This was just the feature I was trying to implement until I looked on github.
Updated docs would be highly appreciated

namirsab:
@bluepnume done!

Note: @twig-openlearning this is happening probably not only to you, but to many people. That's why I started docs as I discover things. I really think this project has a lot of value, but without proper docs it's really difficult to start with. So, please, I encourage you to do the same!

bluepnume:
Awesome, thanks!

PR #124, opened 2017-10-18
add Component.init api doc

freshfunkee:
Lost a few hours trying to dig this function out. I thought it might be useful for anyone who is looking for similar functionality.


bluepnume:
Looks good, thanks!

Issue #131, opened 2017-11-09
Component init() overwrites the context argument


freshfunkee:
https://github.com/krakenjs/xcomponent/blob/9298cfc6735e0f8794abd6ee0acfe3d4c4a2d702/src/component/component/index.js#L413

Calling `init()` and passing a context as the second argument has no effect. The context is always overwritten.

PR #132, opened 2017-11-10
use context as getRenderContext argument

freshfunkee:
Fix for issue #131.

`getRenderContext()` seemed to be used inconsistently. If an element reference, or any value, was passed, `CONTEXT_TYPE.IFRAME` was validated, regardless of whether the component definition specified a context.

It looks like a work around was put in place for `Component` functions like `renderIframe` and `renderPopup`, where `validateRenderContext(context)` was called directly, and the `CONTEXT_TYPE` was hard-coded to the `ParentComponent` constructor arguments. I refactored these functions to be more consistent, so validation is handled correctly by `getRenderContext(context)`

### DIFF
--- a/src/component/component/index.js
+++ b/src/component/component/index.js
@@ -410,8 +410,7 @@ export class Component<P> extends BaseComponent<P> {
     */
 
     init(props : (PropsType & P), context : ?string, element : ElementRefType) : ParentComponent<P> {
-        context = this.getRenderContext(element);
-        return new ParentComponent(this, context, { props });
+        return new ParentComponent(this, this.getRenderContext(context), { props });
     }
 
 
@@ -419,23 +418,39 @@ export class Component<P> extends BaseComponent<P> {
         return new DelegateComponent(this, source, options);
     }
 
-    validateRenderContext(context : string) {
-        if (!this.contexts[context]) {
+    validateRenderContext(context : ?string, element : ?(ElementRefType)) {
+        if (context && !this.contexts[context]) {
             throw new Error(`[${this.tag}] Can not render to ${context}`);
+        }
+
+        if (element) {
+            const defaultContext = this.getDefaultContext();
+
+            if (context === CONTEXT_TYPES.POPUP || defaultContext === CONTEXT_TYPES.POPUP) {
+                throw new Error(`[${this.tag}] Context type ${CONTEXT_TYPES.POPUP} does not support use of an element selector`);
+            }
         }
     }
 
-    getRenderContext(element : ElementRefType) : string {
-
-        if (element) {
-            this.validateRenderContext(CONTEXT_TYPES.IFRAME);
-            return CONTEXT_TYPES.IFRAME;
-        } else if (this.defaultContext) {
+    getDefaultContext() : string {
+        if (this.defaultContext) {
             return this.defaultContext;
         } else if (this.contexts[CONTEXT_TYPES.IFRAME]) {
             return CONTEXT_TYPES.IFRAME;
         } else if (this.contexts[CONTEXT_TYPES.POPUP]) {
             return CONTEXT_TYPES.POPUP;
+        }
+    }
+
+    getRenderContext(context : ?string, element : ?(ElementRefType)) : string {
+        const defaultContext = this.getDefaultContext();
+
+        if (context || element) {
+            this.validateRenderContext(context, element);
+            
+            return context ? context : defaultContext;
+        } else if (defaultContext) {
+            return defaultContext;
         }
 
         throw new Error(`[${this.tag}] No context options available for render`);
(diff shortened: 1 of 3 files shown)


bluepnume:
Hey thanks for the PR! I'll try to get some comments your way and get this merged early next week.

src/component/component/index.js
```diff
@@ -425,11 +424,11 @@ export class Component<P> extends BaseComponent<P> {
         }
     }
 
-    getRenderContext(element : ElementRefType) : string {
+    getRenderContext(context : ?string) : string {
 
-        if (element) {
```
bluepnume: Is there a reason we're removing this check? If `element` is passed and both `iframe` and `popup` are availlable, we should always pick `iframe`

test/tests/error.js
```diff
@@ -72,13 +72,6 @@ describe('xcomponent error cases', () => {
         });
     });
 
-    it('should try to render a component with a specified element when iframe mode is not supported', done => {
```
bluepnume: Would ideally like to keep this test for the above reason.

bluepnume:
Can you remove /dist changes from this PR? That way the git history stays clean/separated, and these changes will get picked up in the next release. Thanks!

src/component/component/index.js
freshfunkee: I think the scenario you have mentioned may be redundant. When we remove the element check, this function will return `iframe`, if both `iframe` and `popup` are defined anyway, due to the following `if/else` statements. With the element check, if `iframe` is not a supported context, but `popup` is, an error will be thrown, even if the component has a valid `defaultContext` of `popup`, or the `init` function was called with `popup` as its context 

The problem with the element check is that this function was always validating `iframe`, regardless of how the component is configured for contexts, or what `defaultContext` is passed. This function was bypassed in favour of directly calling `validateRenderContext` in the `renderIframe` and `renderPopup` functions when instantiating a `ParentComponent`. 

I'm replacing the `element` check with a `context` check. If `iframe` and `popup` are available, and no default is defined, it will use `iframe` anyway as it is the first context to be checked. It makes the usage of this function more consistent throughout the code, and fixes the issue I've referenced with this pull request

test/tests/error.js
freshfunkee: I feel this highlights the issue I have with this function. If an component does not support `iframe`, but does support `popup`, why should an error be thrown if a context isn't specified? Should is not default to `popup` in that case?

src/component/component/index.js
bluepnume: Makes sense. What about the following case; the component only allows `popup`, but `render()` is called with an element. In this case we should throw an error, right?

freshfunkee: An error may be a bit extreme in this case, I think. If an element reference is specified for a `popup`, it won't cause a `ParentComponent` to fail initialization. Maybe a warning would be more appropriate? I would have thought the `render` function is generic enough to instantiate either `context` with little change to the provided configuration.

bluepnume: If we don't throw an error, we either have to:

1. Render to the element using an iframe, even though iframes are not allowed
2. Render to a popup and ignore the specified element

Neither of these are very intuitive for end-users. So it definitely sounds like an error-out case to me. Could we just pass the element through to `validateRenderContext`?

freshfunkee: That may be one solution, by giving the function 2 arguments.

I have a question though, is there any benefit to having a generic `render()` function in the first place? What functionality does it provide that explicit `renderIframe` and `renderPopup` functions don't have? Is there a particular use case for `render()`? Do you think it would be *more* intuitive to remove `render/renderTo` altogether, in favour of  having the end user call `renderIframe/renderIframeTo` and `renderPopup/renderPopupTo`? I feel it would be easier and clearer to define a more explicit API, rather than handling edge cases of a looser API.

bluepnume: So, I guess if I had to go one direction or the other, I'd rather normalize on having only `render()` and `renderTo()` and allowing them to accept an enum value for the context, as an optional param.

The reason for that is, 90% of the use-cases for xcomponent involve writing a component that only works for one context -- it feels pretty rare to me that a component would be used as both a popup and an iframe. So I'd rather leave that in the hands of the component author rather than the component renderer. `renderIframe` and `renderPopup`, in retrospect, feel like an unnecessarily large api surface area.

Anyway, that's a discussion we can probably take outside this PR, if we need.

freshfunkee: Ok. So for this PR, I'll look into throwing an error if a component is configured for a `popup` context only, and an element reference has been specified

bluepnume: Cool, sounds good 👍 happy to merge after that, everything else looks great

freshfunkee:
Updated: Added error check when `popup` is configured for use, and an element selector is specified. Added unit tests.

bluepnume:
Merged -- thanks a lot for taking the time to fix this! Appreciate it.

PR #134, opened 2017-12-01
fixes #133 extra div wrapper in React 16

### LINKED ISSUE #133: React/Vue drivers wrap component in extra div
Since this div isn't styled it breaks responsive dimensions and auto resize. This is present in the demo at http://krakenjs.com/xcomponent/demo/frameworks/react/index.htm The Angular 2 demo doesn't wrap the component this way. 

I'm using React 16 and the demo is as well, so maybe React 16 introduced this? This is my first attempt at using xcomponent so I apologize if I'm just missing something. 

You can see the div here
image

But it is not included in the code, so it seems to be injected by the React driver
image

kriscarle:


bluepnume:
Doesn't this still leave a residual div hanging around in the dom?

kriscarle:
Oh, yeah it does, but looks like the angular 2 driver does that too. I just tested sticking a `el.parentNode.removeChild(el)` at the end of `componentDidMount()` and that seems to do the trick, should I add that to the PR?

image

bluepnume:
`parent.render(el.parentElement);` -- wouldn't this render the component as the last child of the parent? what if there's an element that's supposed to be after the xcomponent, in the same parent?

kriscarle:
So I looked at this a bit more and it looks like the Vue driver demo has the same problem as React. Angular2 is injecting it at the end of the parent, leaving an unused div, but at least parent is the Angular component tag itself. In the case of React and Vue the driver needs to create the container to render xcomponent into, but unlike the basic example, it isn't a user created container and it sits between xcomponent and what the user expects to be the parent. 

I no longer think this PR is the correct way to solve this. I agree, calling render on the parent element is a bad idea and could insert the component in the wrong place. React/Vue are supposed to be managing that...

Maybe `Component.driver()` should allow the user to pass in props so they can manually style the generated container if needed? I've written my own React component wrapper to do that for now instead of using the driver. 

```js
import React from 'react';

export default class XComponentReact extends React.Component<any, void> {
  render() {
    return (
      <div  
        ref={(container) => { this.xcontainer = container; }} 
        {...this.props.containerProps}
      />
    );
  }

  componentDidMount() {
    // eslint-disable-next-line global-require (needed for SSR)
    const xcomponent = require('xcomponent');
    const xEmbedComponent = xcomponent.create(this.props);
    xEmbedComponent.log(`instantiate_react_component`);

    let parent = xEmbedComponent.init(Object.assign({}, this.props), null, this.xcontainer);
    this.setState({parent});
    parent.render(this.xcontainer);
  }

  componentDidUpdate() {
    if (this.state && this.state.parent) {
      this.state.parent.updateProps(Object.assign({}, this.props));
    }
  }
}
```

bluepnume:
Yeah I definitely need to mull this over some more. It would be great if React etc. let you directly return a dom node in `render()`. I'd love to find a way to do this which doesn't leave any room for breaking down the line if any of these frameworks change how they work.

PR #147, opened 2018-03-20
Reference container when rendering MyLoginXComponent in lightbox demo

### LINKED ISSUE #146: xcomponent lightbox demo appears to be broken
Tried visiting http://krakenjs.com/xcomponent/demo/basic/lightbox/index.htm on Chrome and Safari, both fail to render the lightbox and show the following error:

`index.js:451 Uncaught Error: [my-login-component] Context type iframe requires an element selector`

Seems like there's a missing reference to `'#container'` here: https://github.com/krakenjs/xcomponent/blob/85d14e1a4f2a41c3aa93e29cb6adeef6b451e480/demo/basic/lightbox/index.htm#L32

Screenshot of the error, using `Chrome OSX Version 65.0.3325.181 (Official Build) (64-bit)`

xcomponent-lightbox-screenshot

mikegrassotti:
Fixes #146

### DIFF
--- a/demo/basic/lightbox/index.htm
+++ b/demo/basic/lightbox/index.htm
@@ -29,7 +29,7 @@
                 document.querySelector('#result').innerText = email + ' logged in!';
             }
 
-        });
+        }, '#container');
 
     </script>
 


bluepnume:
Can we make this render into `'body'`? Ideally lightbox-style components should be rendered directly into the body so they can show a background overlay over the whole page.

Thanks for taking the time! :)

PR #148, opened 2018-03-20
Improve autoResize description with element property

jelohe:
The docs doesn't explain how to override the `element` property on the `autoResize` section.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -412,7 +412,7 @@ defaultLogLevel: 'error'
 Note that this value can be overriden by passing 'logLevel' as a prop when rendering the component.
 
 
-#### autoResize `boolean | Object<String, Boolean>`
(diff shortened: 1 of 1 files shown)


docs/api.md
```diff
@@ -412,7 +412,7 @@ defaultLogLevel: 'error'
 Note that this value can be overriden by passing 'logLevel' as a prop when rendering the component.
 
 
-#### autoResize `boolean | Object<String, Boolean>`
+#### autoResize `boolean | Object<String, Boolean | String>`
```
bluepnume: Can we change this to `boolean | { height : boolean, width : boolean, element : string }`

bluepnume:
Thanks! Just one minor comment then will gladly merge.

jelohe:
Done, good catch!

PR #150, opened 2018-03-27
Make instructions for quick start example more explicit

crookedneighbor:
Trying to implement this, it took me a while to figure out that I needed to define `MyLoginComponent` on both pages. This isn't stated explicitly in the quick start.

This PR makes it clearer and easier to copy paste for a working example.


bluepnume:
Thanks a mil @crookedneighbor!

PR #153, opened 2018-04-19
Make it optional to include IE bridge support when importing as module

r0stig:
Solves: https://github.com/krakenjs/xcomponent/issues/152

When included by webpack you can choose if you want to include with or without bridge support the same way react does it (https://github.com/facebook/react/blob/master/packages/react/npm/index.js):

```
plugins: [
    new webpack.DefinePlugin({
      'process.env': {
        'NODE_ENV': JSON.stringify(process.env.NODE_ENV),
        'XCOMPONENT_EXCLUDE_IE_BRIDGE': true
      }
    })
  ]
```

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,5 +1,5 @@
 
-if (process.env.XCOMPONENT_EXCLUDE_IE_BRIDGE) {
+if (process.env.XCOMPONENT_FRAME_ONLY) {
   module.exports = require('./dist/xcomponent.frame');
   module.exports.default = module.exports;
 } else {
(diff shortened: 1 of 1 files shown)


index.js
```diff
@@ -1,3 +1,8 @@
 
-module.exports = require('./dist/xcomponent');
-module.exports.default = module.exports;
+if (process.env.XCOMPONENT_EXCLUDE_IE_BRIDGE) {
```
bluepnume: Can we make this `process.env.XCOMPONENT_FRAME_ONLY` for consistency? (since it has ramifications beyond just IE)

PR #155, opened 2018-05-05
Refactored Vue driver and tests to remove dependency on template compiler

doublemarked:
The Vue driver is quite simple but was using string templates, which require an additional dependency of the Vue template compiler (~20Kb gzipped). I've refactored the driver to use a render function. I've also refactored the Vue tests to use render functions, in order to entirely eliminate the template compiler from the testing stack. 

Please let me know your thoughts and if you would like any changes to this PR! 

Documentation regarding Vue builds (runtime vs. runtime+compiler): https://vuejs.org/v2/guide/installation.html#Runtime-Compiler-vs-Runtime-only


bluepnume:
Thanks for the PR! looks good to me

PR #168, opened 2018-06-26
(fix) dimesions needs to be px or %

joshgreenwell:
Create does validation on the dimensions object to make sure the values entered are a px or % string, if they are not an error is thrown


bluepnume:
Thanks!

PR #198, opened 2018-12-08
update examples of containerTemplate api docs

daton89:
I updated the examples in docs/api.md of containerTemplate adding node el for outlet container


bluepnume:
Thanks!

PR #224, opened 2019-04-02
Update React example to use Component.driver syntax

copperwall:
This updates the React example snippet to match the example on `docs/api.md`. I don't believe that Zoid has a `react` property on the Component instance anymore, so this updates it to use `Component.driver` and specifies react as the framework.


bluepnume:
Thanks!

Issue #241, opened 2019-06-10
Documentation around `containerTemplate` and events


mdalzell:
Currently the documentation around `containerTemplate` can be confusing to new users (it was for me).  The examples given don't reproduce behavior similar to the `defaultContainerTemplate`, and omit some details that would probably be necessary for any production use of this feature.

A few suggestions to improve:
* Specify what is in the `opts` parameter.  Most importantly, `event` is not in the documentation, and that is crucial to swapping the prerender frame out for the main one.
* Document what `props` are passed in.  For example, `cspNonce` is used in the default template, some documentation around this would be appreciated.
* Document the different types of events and when they fire.  I saw there was another question around `RESIZE` that probably could've been answered by this documentation.
* Ideally, the examples given would reproduce the behavior of the default template.  Alternatively, it might be good to include the default template right there as a more robust sample, instead of having to dig into the source.

I'd be happy to help out with updating this, however I'm still new to this library so I might need some guidance on some of this.

mdalzell:
#243

bluepnume:
Thanks for taking care of this one!

PR #243, opened 2019-06-15
#241 Update API doc with event info for templates

mdalzell:
Issue 241: Updated the documentation around containerTemplate and prerenderTemplate to include information on events.  Also clarified that the opts parameter is the same for both functions, and included a link to view the default containerTemplate.


bluepnume:
Thanks!

Issue #202, opened 2019-01-15
The spinner just keeps going by default


lolcode8:
The spinner just keeps going by default. The page seems to render 2 Iframes
The one I want is hidden (opacity: 0) and the one with the spinner has opacity: 1

Why is this happening?
<img width="993" alt="screen shot 2019-01-14 at 9 31 44 pm" src="https://user-images.githubusercontent.com/37151383/51154996-fabbd500-1843-11e9-8db5-4ec4175edc86.png">
<img width="954" alt="screen shot 2019-01-14 at 9 31 59 pm" src="https://user-images.githubusercontent.com/37151383/51154997-fb546b80-1843-11e9-939c-8d986fdb0f38.png">

sytong:
Have you included the compiled JS script (`browser-zoid-component.js`??) in the HTML rendered in `http://localhost:8080/search-real-estate-map?ON`?

bluepnume:
Do you still have this issue? Can you share a page where we can repro?

flexzuu:
The react demo seems to have the same problem
http://krakenjs.com/zoid/demo/frameworks/react/index.htm 

In general seems you made breaking changes that are not yet documented. If you need help getting docs up with changes let me know.

bluepnume:
Fixed framework demos with https://github.com/krakenjs/zoid/commit/eed13c35ffebef17425628cc1af90a2572f2510c

r0stig:
I just upgraded from the 7.x branch to the 9.x branch and found out that you have to include the prerenderFrame in the output of containerTemplate. In my use case I don't want to show anything when loading the iframe and are returning `null` in the prerenderTemplate-function (which was enough in the 7.x-branch). To solve this I wrapped the prerenderFrame within a container which sets `display: none`.

I'd wish for a simpler way to render a component without the preload-iframe if possible.

muhaimincs:
> Have you included the compiled JS script (`browser-zoid-component.js`??) in the HTML rendered in `http://localhost:8080/search-real-estate-map?ON`?

how to import it?

bluepnume:
> I just upgraded from the 7.x branch to the 9.x branch and found out that you have to include the prerenderFrame in the output of containerTemplate. 

Really? What happens if you don't include it?

That sounds like a bug. If you can provide some details, please open up a new issue and we'll get to the bottom of it.

PR #255, opened 2019-08-12
Update documentation to avoid scheme-relative URLs

bpartridge:
Per the HTML spec you can put a scheme-relative url, e.g. `//google.com` as the src of an iframe: https://html.spec.whatwg.org/multipage/iframe-embed-object.html#attr-iframe-src

But Zoid sniffs the protocol in isSameDomain (from cross-domain-utils) so it expects http or https. If a protocol is not provided, the preload page spinner spins infinitely as in #202.

This makes it explicit that a protocol should be provided.

### DIFF
--- a/docs/api.md
+++ b/docs/api.md
@@ -18,7 +18,9 @@ tag: 'my-component-tag'
 
 #### url `string | ({ props }) => string` [required]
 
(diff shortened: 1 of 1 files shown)


bluepnume:
Thanks for this!

PR #256, opened 2019-08-16
Allow tag names without dashes

ctavan:
Is there a specific reason to enforce a dash in the tag name? I stumbled upon this one while upgrading from zoid v6 (where dashless tag names were still working) to v9 and it came pretty unexpected.

I've added a test-case to cover my change however I'm not sure whether the place where I put it is right because it's the first test in that file that doesn't expect an error.

If you have an idea where to put the test case instead just let me know and I'll amend my PR.


bluepnume:
Looks good, thank you!

PR #307, opened 2020-06-01
fix(): use console warning for zoid version mismatch

### LINKED ISSUE #225: Suggestion: can you turn off the parent/child version check error?
We did some minor upgrade on our Zoid widgets and deployed, then we got a series of above errors. I think because we have cached the child scripts and thus it caused the mismatch.

The caching technique is pretty common when we developed client sources and maybe it not worth it to fail the widget entirely just because of the minor x.x.XXX number changes?

please help us, it is critical

blumk:
Fixes #225 

For various reasons it's not always possible to deploy the parent / child at the same time.

Currently zoid crashes on a version mismatch even if there are no breaking changes between two versions.

Rationale for using `console.warn()` instead of `new Error()`:
- In case of a breaking change the software still crashes with a warning on console
- non-breaking changes allow independent deployments (think of a security patch)

Ultimately it's the developers responsibility to make sure that the software is working as intended and that there are no console warnings.

### DIFF
--- a/src/child/child.js
+++ b/src/child/child.js
@@ -122,7 +122,7 @@ export function childComponent<P>(options : NormalizedComponentOptionsType<P>) :
     }
 
     if (childPayload.version !== __ZOID__.__VERSION__) {
-        throw new Error(`Parent window has zoid version ${ childPayload.version }, child window has version ${ __ZOID__.__VERSION__ }`);
+        console.warn(`Zoid version mismatch detected. Parent window has zoid version ${ childPayload.version }, child window has version ${ __ZOID__.__VERSION__ }`);
     }
 
     const { parent: parentRef, parentDomain, exports, context, props: propsRef } = childPayload;


songz:
@bluepnume any reason why this wasn't merged? Does PayPal want to intentionally throw error on version mismatch?

bluepnume:
@songz -- it's a choice between:

a) Fail at this point with a clear error message
b) Fail at some future indeterminate point due to some unforeseen incompatibility between the messaging protocol in two different zoid versions (e.g. some new key expected in payloads, etc.)

We were hitting b) a lot. So a) at least gives a super clear reason.

Typically with zoid, the expectation is you create a CDN script with your component + the zoid library, and make sure you're loading the exact same CDN script in the parent window and in any child frames. That avoids any such compatibility issues, and it also has performance benefits (you usually don't want to force your user to download multiple versions of zoid).

^ We should probably document the above a little better.

akshay-8d66:
> Typically with zoid, the expectation is you create a CDN script with your component + the zoid library, and make sure you're loading the exact same CDN script in the parent window and in any child frames. T

Could you point out where `paypal/paypal-checkout-components` does this.

Issue #308, opened 2020-06-04
Attributes, based on props


OlehRb:
In our widget we are using user's location and it seems, like simple
```
...
attributes: {
  iframe: {
    allow: 'geolocation',
  },
},
...
```
doesn't work anymore as Features Policy now requires the iframe's url to be passed explicitly, like: `allow: 'geolocation https://my-awesome-widget.com'`

The problem we faced is that we have different environments: development, stage and production and they all have different urls

The environment is passed as an optional prop, but it seems like there is no way to define attributes, based on props

However, it seems, like we can define iframe's attributes based on props, if we use custom `containerTemplate`, but in this approach we would have to either lose or reimplement / copy-paste the functionality of autoresize and visible / hidden classes switching, which are embedded to the `defaultContainerTemplate`

So, is it possible to either define iframe's attributes, based on props or in some way extend the `defaultContainerTemplate`, so instead of copy-pasting whole default template I could do something, like:

```
...
containerTemplate: (options) => {
  // urls here is just a map, where environment is a key and url is a value
  options.frame.allow = `geolocation ${urls[options.props.environment]}`;

  return defaultContainerTemplate(options);
}
...
```

Any help would be much appreciated

bluepnume:
This is a fair ask. We can probably allow `attributes` to be a function that takes props. If you'd like to make a PR with that change, I'd be happy to merge it in 👍

OlehRb:
@bluepnume, thanks, I will give it a try, most probably during this weekend, and get back to you as soon as I have something :+1:

OlehRb:
@bluepnume, I've submitted a PR https://github.com/krakenjs/zoid/pull/311
It seems to be rather a small change actually

Can you, please, take a look at it?

OlehRb:
@bluepnume , ping

OlehRb:
@bluepnume , sorry to bother you, but when do you expect the version this version to be released to npm?

PR #311, opened 2020-06-12
Props based attributes

OlehRb:
Resolves https://github.com/krakenjs/zoid/issues/308 via adding a possibility to define `attributes` field in the same way as we do for `url` field

### DIFF
--- a/src/component/component.js
+++ b/src/component/component.js
@@ -89,7 +89,7 @@ export type NormalizedComponentOptionsType<P> = {|
 
     allowedParentDomains : StringMatcherType,
 
-    attributes : AttributesType,
+    attributes : AttributesType | ({| props : PropsType<P> |}) => AttributesType,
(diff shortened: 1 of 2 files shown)


bluepnume:
Code looks great, thanks for adding this! Would you mind adding a quick test to make sure attributes are set correctly in both cases? Thanks!

OlehRb:
@bluepnume , it seems like my original approach does not work for dynamic attributes, because it looks like the `props` variable gets mutated somewhere along the way..

So I've updated the computation of dynamic attributes to be performed in a dedicated function (the same way as it is implemented for `url` parameter)

Also I've added a few tests for dynamic and static attributes
Please, let me know if you have any concerns about current implementation or if I can improve / extend test cases in any way (cause to be honest, I don't have too much of experience with tests).

PR #314, opened 2020-06-25
feat(bridge): allow bridgeUrl to be a function

louietan:
`url` and `bridgeUrl` are on the same domain, it would be nice for `bridgeUrl` to also accepts a function so that we can have different bridge URLs for different environments.


bluepnume:
Awesome, thanks!

PR #325, opened 2020-10-08
closes #291 - make sure process exists before checking process.env an…

### LINKED ISSUE #291: process is assumed to exist, which it doesn't in the browser
When trying to use a package (which depends on zoid) we get an error that points to:
`https://github.com/krakenjs/zoid/blob/bad7235289672fe374c8f1fd5b000918ff9193e3/index.js`

Which is assuming that `process` exists (which it doesn't in the browser).
If fixed we can remove this polyfill:
```
(window as any).process = {
    env: { DEBUG: undefined },
};
```

Environment Angular 9

anthonykhoa:
## The issue
#291 
The `index.js` file has a problem when it is being used in an environment that does not use `process`.

## The solution
We can check if `process` exists before checking if `process.env` exists, before checking if `process.env.ZOID_FRAME_ONLY` exists

## This PR will
* Make sure `process` and `process.env` exists before checking if `process.env.ZOID_FRAME_ONLY` exists using the `logical AND` operator `&&`

## Other ways to check for existence of child keys in an object
* Use lodash library's [._get function](https://lodash.com/docs/4.17.15#get).
    * This repo does not use lodash, so we should not use lodash just for this issue.
* Use [optional chaining](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Optional_chaining)
    * This is not as readable.

### DIFF
--- a/index.js
+++ b/index.js
@@ -4,12 +4,12 @@
 // eslint-disable-next-line no-process-env
 if (process && process.env && process.env.ZOID_FRAME_ONLY) {
     // $FlowFixMe
-    module.exports = require("./dist/zoid.frame");
+    module.exports = require('./dist/zoid.frame');
     // $FlowFixMe
     module.exports.default = module.exports;
 } else {
     // $FlowFixMe
-    module.exports = require("./dist/zoid");
+    module.exports = require('./dist/zoid');
     // $FlowFixMe
     module.exports.default = module.exports;
 }
--- a/index.js
+++ b/index.js
@@ -2,14 +2,14 @@
 /* eslint import/no-commonjs: 0 */
 
 // eslint-disable-next-line no-process-env
-if (process.env.ZOID_FRAME_ONLY) {
+if (process && process.env && process.env.ZOID_FRAME_ONLY) {
     // $FlowFixMe
-    module.exports = require('./dist/zoid.frame');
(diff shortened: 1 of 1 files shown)


songz:
@mstuart @bluepnume

mnicpt (approved):

mnicpt:
Thanks, @anthonykhoa !

index.js
```diff
@@ -2,7 +2,7 @@
 /* eslint import/no-commonjs: 0 */
 
 // eslint-disable-next-line no-process-env
-if (process.env.ZOID_FRAME_ONLY) {
+if (process && process.env && process.env.ZOID_FRAME_ONLY) {
```
sibiraj-sr, line 5: this will still throw error since process is not defined in browser.

``` js
if (typeof process !== 'undefined' && process.env && process.env.ZOID_FRAME_ONLY) {
```

ar7casper:
@mnicpt Idea when it's gonna be released?

mnicpt:
@ar7casper I will look to see if we can release tomorrow.

ar7casper:
BTW, I found the issue I had.
Writing it down in case someone gets here in the future -
I used webpack with Dotenv plugin.
Once I switched to webpack.DefinePlugin it worked and the fix is not needed.

PR #329, opened 2020-10-09
closes #328 - Update .travis.yml file to fix CI build failure issue

### LINKED ISSUE #328: Travis CI builds are failing
The Travis CI builds are failing, with the error being 
Screen Shot 2020-10-08 at 5 24 37 PM

anthonykhoa:
### The issue
#328 
The Travis CI builds are failing, with the error being 
Screen Shot 2020-10-08 at 5 24 37 PM

What was happening was that older versions of installed modules were being used. This solution to this problem is to remove the node_modules folder before travis CI starts installing dependencies.

### This PR will
* Update the `.travis.yml` file to remove `node_modules` folder before running tests

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,4 +1,6 @@
 language: node_js
 node_js:
-   - "lts/*"
+  - 'lts/*'
+before_install:
+  - rm -rf node_modules
 script: npm test


mstuart:
Thanks @anthonykhoa! I don't think we realized the build was broken. We'll add a PR status check that'll prevent merging until the build passes from now on :) ty!! 🙏

PR #346, opened 2021-05-12
render within shadow DOM test added

gabrielo91:
# Description
- Adde new test for nested shadow DOM support

## Nested shadow DOM support PR

- https://github.com/krakenjs/belter/pull/54

### DIFF
--- a/test/common.js
+++ b/test/common.js
@@ -164,7 +164,7 @@ export function runOnClick<T>(handler : () => T) : T {
     }
 }
 
-export function getContainer({ parent, shadow = false, slots = false } : {| parent? : ?HTMLElement, shadow? : boolean, slots? : boolean |} = {}) : {| container : HTMLElement, destroy : () => void |} {
+export function getContainer({ parent, shadow = false, slots = false, nested = false } : {| parent? : ?HTMLElement, shadow? : boolean, slots? : boolean, nested? : boolean |} = {}) : {| container : HTMLElement, destroy : () => void |} {
     const parentContainer = parent = parent || document.body;
     
     if (!parentContainer) {
@@ -183,6 +183,7 @@ export function getContainer({ parent, shadow = false, slots = false } : {| pare
 
     const customElementName = `zoid-custom-element-${ uniqueID() }`;
     const customSlotName = `zoid-custom-slot-${ uniqueID() }`;
+    const innerWrapperName = `zoid-inner-wrapper-${ uniqueID() }`;
     
     const shadowContainer = document.createElement(slots ? 'slot' : 'div');
     shadowContainer.setAttribute('name', customSlotName);
@@ -201,6 +202,13 @@ export function getContainer({ parent, shadow = false, slots = false } : {| pare
         }
     });
 
+    customElements.define(innerWrapperName, class extends HTMLElement {
+        connectedCallback() {
+            const shadowRoot = this.attachShadow({ mode: 'open' });
+            shadowRoot.appendChild(shadowContainer);
+        }
+    });
+
     const customElement = document.createElement(customElementName);
     parentContainer.appendChild(customElement);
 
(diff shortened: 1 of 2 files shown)


bluepnume:
Looks great! One thing I don't understand -- how is the test passing before https://github.com/krakenjs/belter/pull/54 is merged?

gabrielo91:
> Looks great! One thing I don't understand -- how is the test passing before [krakenjs/belter#54](https://github.com/krakenjs/belter/pull/54) is merged?

Still wondering the same, today later we can discuss the approach 😅

test/tests/happy.jsx
```diff
@@ -440,6 +440,34 @@ describe('zoid happy cases', () => {
...
+                });
+            };
+
+            onWindowOpen().then(expect('onWindowOpen', ({ win }) => {
+                if (getParent(win) !== window) {
+                    throw new Error(`Expected window parent to be current window`);
+                }
+            }));
+
+            const { container, destroy } = getContainer({ shadow: true, nested: true });
```
gregjopa, line 460: @gabrielo91 and I were pairing on this and noticed that `slot: true` is missing. Adding it here helps the new nested use case to get hit. 

```suggestion
            const { container, destroy } = getContainer({ shadow: true, nested: true, slot: true });
```

test/common.js
```diff
@@ -211,6 +219,32 @@ export function getContainer({ parent, shadow = false, slots = false } : {| pare
         };
     }
 
+    if (nested) {
+        const innerWrapper = document.createElement('inner-element-wrapper');
+        innerWrapper.setAttribute('id', 'inner-element-wrapper-id');
+
+        const customElementShadowRoot = customElement.shadowRoot;
```
gregjopa, line 226: There is still some logic here that isn't working as expected. @gabrielo91 will continue to debug.

bluepnume:
This is the first time I've ever commented on a PR to say "the tests are passing! please fix the PR so the tests break!" :D

test/common.js
```diff
@@ -201,9 +202,43 @@ export function getContainer({ parent, shadow = false, slots = false } : {| pare
...
+
     const customElement = document.createElement(customElementName);
     parentContainer.appendChild(customElement);
 
+    if (nested) {
+        const innerWrapper = document.createElement(innerWrapperName);
+        innerWrapper.setAttribute('id', 'inner-element-wrapper-id');
+
+        const customElementShadowRoot = customElement.shadowRoot;
+        const innerWrapperShadowRoot = innerWrapper.shadowRoot;
```
gregjopa, line 220: The `innerWrapper` isn't in the DOM so we can't get at it's `shadowRoot` yet. Please try this instead:

```js
if (nested) {
        const innerWrapper = document.createElement(innerWrapperName);
        innerWrapper.setAttribute('id', 'inner-element-wrapper-id');

        const customElementShadowRoot = customElement.shadowRoot;

        if (customElementShadowRoot) {
            customElementShadowRoot.appendChild(innerWrapper);
        }

        const innerWrapperShadowRoot = innerWrapper.shadowRoot;
        const innerWrapperContainer = document.createElement('div');
        innerWrapperContainer.setAttribute('id', 'super-id');

        if (innerWrapperShadowRoot) {
            innerWrapperShadowRoot.appendChild(innerWrapperContainer);
        }

        return {
            container: innerWrapperContainer,
            destroy:   () => {
                parentContainer.removeChild(customElement);
                parentContainer.removeChild(innerWrapper);
            }
        };
    }
```

gregjopa: After this code change we should expect to get the following error message `Host element is also in shadow dom`:

expected-failure-message

gregjopa (approved):
Looks great! We are getting the expected error now. Upgrading to the latest belter should fix this test failure.

bluepnume (approved):
LGTM! Thanks Gabriel. This should give a lot of confidence for any component using zoid with shadow dom.

PR #351, opened 2021-06-02
Vue driver support for kebab-case

leogedler:
**Added support to handle kebab-case props in the Vue driver**

According Vue style guide best practices the props on template components should be passed in in kebab-case format.

_Vue docs reference_: https://vuejs.org/v2/style-guide/#Prop-name-casing-strongly-recommended.
_Issue reference_: https://github.com/paypal/paypal-checkout-components/issues/1338#issuecomment-777005173
_Associated PR_: https://github.com/paypal/paypal-checkout-components/pull/1650

### DIFF
--- a/src/drivers/vue.js
+++ b/src/drivers/vue.js
@@ -1,6 +1,6 @@
 /* @flow */
 
-import { noop } from 'belter/src';
+import { noop, dasherizeToCamel } from 'belter/src';
 
 import type { ComponentDriverType } from '../component';
 import { CONTEXT } from '../constants';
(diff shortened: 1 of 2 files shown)


src/drivers/vue.js
```diff
@@ -25,6 +25,16 @@ type VueType = {|
     component : (string, VueComponent) => RegisteredVueComponent
 |};
 
+function propsToCamelCase(props : Object) : Object {
+    for (const key of Object.keys(props)) {
```
gregjopa, line 29: Can we avoid mutating the props object passed in and instead return a new object? For example, we could use `Array.prototype.reduce` to return a new altered copy: Ex:

```js
function propsToCamelCase(props : Object) : Object {
  return Object.keys(props).reduce((key, acc) => {
            // add logic here
  }, {})
}
```

leogedler: @gregjopa I have updated the logic to use reduce instead of mutating the object as suggested.

gregjopa (approved):

PR #357, opened 2021-08-03
Vue driver logic for prop name replacement (style prop).

leogedler:
**Reasoning**

In Vue the prop name `style` is reserved thus if the user wants to pass in a style for the PayPal button this won't work, instead the user now can pass in a `style-object` prop with the desire styling for the button.
To keep Zoid generic, we are passing in the original prop name, so if Zoid is used for other application the user will received the expected prop name in the component.

**Change**
Added logic on the Vue driver to replace the prop name `styleObject` or `style-object` by a `style` prop.

Reference: https://vuejs.org/v2/guide/class-and-style.html#Object-Syntax-1

### DIFF
--- a/src/drivers/vue.js
+++ b/src/drivers/vue.js
@@ -28,7 +28,12 @@ type VueType = {|
 function propsToCamelCase(props : Object) : Object {
     return Object.keys(props).reduce((acc, key) => {
         const value = props[key];
-        if (key.includes('-')) {
+        // In vue `style` is a reserved prop name
+        if (key === 'style-object' || key === 'styleObject') {
+            acc.style = value;
+            // To keep zoid as generic as possible, passing in the original prop name as well
+            acc[key] = value;
+        } else if (key.includes('-')) {
             acc[dasherizeToCamel(key)] = value;
         } else {
             acc[key] = value;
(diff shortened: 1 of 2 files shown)


gregjopa (approved):
Nice work!

mstuart:
lgtm!

westeezy (approved):

src/drivers/vue.js
```diff
@@ -28,7 +28,12 @@ type VueType = {|
 function propsToCamelCase(props : Object) : Object {
     return Object.keys(props).reduce((acc, key) => {
         const value = props[key];
-        if (key.includes('-')) {
+        // In vue `style` is a reserved prop name
+        if (key === 'style-object' || key === 'styleObject') {
+            acc.style = value;
+            // To keep zoid as generic as possible, passing in the original prop name as well
+            acc[key] = value;
```
andregardi, line 35: If key is indeed `style-object`, this would not convert it to camel case, right?
Do we want that? If not, using `acc.styleObject = value` would force the camel case even on this scenario.

leogedler: @andregardi you are right, here is better just set:

```
acc.styleObject = value;
```

I have done the change.

andregardi: Great! Thank you for that!

andregardi (approved):

PR #358, opened 2021-08-03
Fixing broken demos for basic popup and missing script demo

sioked:
While doing some reading of docs I noticed a couple non-functional demos. The popup demo is an easy fix, but I couldn't find a reference to the "script tag component" so just deleted that.

Not sure how to get this published to the krakenjs docs site, will need help with that.

### DIFF
--- a/demo/basic/popup/index.htm
+++ b/demo/basic/popup/index.htm
@@ -4,7 +4,7 @@
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-material-design/4.0.2/bootstrap-material-design.css" />
     <link rel="stylesheet" href="../../common/index.css" />
 
-    <script src="../../../node_modules/jsx-pragmatic/dist/jsx-pragmatic.js"></script>
+    <script src="https://unpkg.com/jsx-pragmatic@2.0.20/dist/jsx-pragmatic.js"></script>
     <script src="../../../dist/zoid.js"></script>
     <script src="./login.js"></script>
 </head>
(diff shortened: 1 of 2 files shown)


demo/basic/popup/index.htm
```diff
@@ -4,7 +4,7 @@
     <link rel="stylesheet" href="https://cdnjs.cloudflare.com/ajax/libs/bootstrap-material-design/4.0.2/bootstrap-material-design.css" />
     <link rel="stylesheet" href="../../common/index.css" />
 
-    <script src="../../../node_modules/jsx-pragmatic/dist/jsx-pragmatic.js"></script>
+    <script src="https://npmcdn.com/jsx-pragmatic@2.0.20/dist/jsx-pragmatic.js"></script>
```
gregjopa, line 7: At first I was going to recommend using `unpkg.com` but then I clicked on this link and realized it redirects to unpkg. TIL 😄

gregjopa (approved):
This seems like a great use case for using unpkg/npmcdn urls. Nice work!

demo/basic/popup/index.htm
sioked: Although they're the same, npmcdn was the older url for unpkg and it looks like they prefer folks to use the new url. I'm pushing up a small update to use that instead.

sioked:
@gregjopa @bluepnume I don't have write access to this repo, nor do I know how to deploy these docs updates. If either of you gets a chance, can you merge the PR in and show me how you deploy?

Thanks!

PR #365, opened 2021-08-17
New drivers for Vue3 and Angular5+

andregardi:
#### Reason
The drivers to register components on frameworks were not working for Angular5+ (starting on version 5.0.0) and Vue3 (starting on version 3.0.0).

#### Solution
- Adapt the current `angular2` driver to work with all versions from `2.0.0` to  the newest `12.2.1`
- Create a new driver for `vue3`. Vue3 has breaking changes in which the components are registered on the `vueApp` instance while on Vue2 it was possible to register it directly on the `Vue` object as a static method.
- It was also necessary to rewrite the way karma load the **Vue** scripts. Because both Vue2 and Vue3 sets the same object name `Vue` on the window object, we can't preload both libs on `karma.conf.js`. As a solution we setup a `loadScript` helper to dynamically load each Vue version, when it is necessary.


src/drivers/angular2.js
```diff
@@ -47,62 +48,74 @@ const equals = (obj1, obj2) => {
...
+                            // $FlowFixMe
+                            return zone.run(() => item.apply(this, arguments));
+                        };
+                    }
+                    return item;
+                });
+            }
+
+            ngOnInit() {
+                const targetElement = this.elementRef.nativeElement;
```
mnicpt, line 83: Should we null check this or provide validation if `elementRef` is not passed?

test/common.js
```diff
@@ -265,3 +265,17 @@ export function getBody(win? : SameDomainWindowType = window) : HTMLBodyElement
 
     return win.document.body;
 }
+
+// eslint-disable-next-line no-restricted-globals, promise/no-native
+export async function loadScript(url : string) : Promise<void> {
+    const scriptelement = document.createElement('script');
```
mnicpt, line 271: Can we camel-case? `scriptElement`

test/common.js
```diff
@@ -265,3 +265,17 @@ export function getBody(win? : SameDomainWindowType = window) : HTMLBodyElement
...
+    scriptelement.setAttribute('src', url);
+    getBody().prepend(scriptelement);
+
+    // eslint-disable-next-line no-restricted-globals, promise/no-native
+    const scriptPromise = new Promise((resolve) => {
+        scriptelement.addEventListener('load', () => resolve());
+    });
+
+    await scriptPromise;
```
mnicpt, line 280: Why does `scriptPromise` need to be a Promise?  We will only resolve on load of script.  Also, what about loading error?

mnicpt:
@andregardi Should we have any new tests for Angular 4+?

test/common.js
andregardi: Definitively!
Thank you for catching that. It is done now.

src/drivers/angular2.js
andregardi: I am not sure if I see a scenario in which `elementRef` is not passed.
The element `elementRef` is injected by the Angular lib itself on our constructor.

Are you seeing something that I am not?
If you think that we need that anyway, just let me know and I will be happy to add it.

test/common.js
andregardi: In our tests, we want to be sure that the script is loaded.
For example, when dynamically loading the vue script, we want to be sure that the `window.Vue` is defined before creating the zoid component.

The `<script>` tags on html usually blocks the javascript executions until the file is loaded. But if it is added at runtime, as we are doing here, the document will not interpret it right away and `window.Vue` will not be available until we await for the `onload` event of this `<script>` tag.

Regarding loading errors, the `loadScript` function is used to load scripts that are served locally by karma itself. Even if it was possible to fail, we already have a native error like:
`GET file:///base/test/lib/vue_v2.5.16.runtime.min.js::ERR_FILE_NOT_FOUND`
But if you want to add a **onerror** event anyway, I will be happy to add it.

andregardi:
> @andregardi Should we have any new tests for Angular 4+?

I am working on that.
The current tests uses Angular 4 lib, even if the driver is called an `angular2`.
I am trying to add a dynamic scripting load, to have a dedicated tests that uses Angular 12 lib, similar of what we have to Vue2 vs Vue3. The value of this tests is to show that the same driver even with the new versions of Angular.

At the moment things a breaking and I am still looking for the reason/solution. But I think we will get that to work soon.

andregardi:
@mnicpt 
I added a test that dynamically loads Angular 12 and use it to test the `angular2` driver.
At same time I kept the old tests as it was. It help us to guarantee that the new driver keeps compatible with the old Angular versions, while it also works with the newest Angular 12.

test/common.js
mnicpt: @andregardi I see what you're doing.  I typically would just return a new Promise and resolve on load of the script rather than initialize a Promise and await on it.  It's in the tests so no biggy.

mnicpt (approved):

test/common.js
andregardi: @mnicpt today I read that again with fresh eyes and it makes complete sense to return the promise instead await it.
I changed that, thank you for pointing that out.

westeezy (approved):
lgtm just a minor comment

gregjopa (approved):

PR #377, opened 2021-11-02
fix: ensure input props are merged in

westeezy:
Bug fix for messaging issue introduced in sdk version: 5.0.268

### Description
`paypal-messaging-components` has a prop def for `account` in which the value is passed in from `inputProps` in some cases and also `propDef.value` may be returning `undefined`. This causes an issue where decorators which read from that `inputProp` will receive the `undefined` value. In our case the decorator in messaging threw on an `undefined` value and that code is linked below.

### Root Cause
During a previous changeset in https://github.com/krakenjs/zoid/pull/376 the `extendProps` function was changed and no longer took into account merging `existingProps` & `inputProps`. Now later in the flow if a decorated prop requires a value from `inputProps` it won't be present.

### Messaging Component's Error
The decorator in `paypal-messaging-components` which received the `undefined` value for account is [here](https://github.com/paypal/paypal-messaging-components/blob/f0cd8d1e9b6dce5ad4602ad559e3f4f4824f10d5/src/zoid/message/component.js#L317). As you can see this decorator uses the `stringStartsWith` from `core-js` which unfortunately will throw on non string values. In this case we were passing in `undefined` as `inputProps` was not properly passed in and it took the `undefined` value from the original propsDef for account.

### DIFF
--- a/src/parent/props.js
+++ b/src/parent/props.js
@@ -11,7 +11,7 @@ import type { ParentHelpers } from './index';
 export function extendProps<P, X>(propsDef : PropsDefinitionType<P, X>, existingProps : PropsType<P>, inputProps : PropsInputType<P>, helpers : ParentHelpers<P>, container : HTMLElement | void) {
     const { state, close, focus, event, onError } = helpers;
 
-    const newProps = { ...existingProps };
+    const newProps = { ...existingProps, ...inputProps };
 
     // $FlowFixMe
     eachProp(inputProps, propsDef, (key, propDef, value) => {
--- a/test/tests/props.js
+++ b/test/tests/props.js
@@ -87,6 +87,53 @@ describe('zoid props cases', () => {
         });
     });
 
+    it('should merge props definition with input props before passing to decorator', () => {
+        return wrapPromise(({ expect }) => {
+            window.__component__ = () => {
+                return zoid.create({
+                    tag:    'test-prop-value-container-url-param-bug',
+                    url:    'mock://www.child.com/base/test/windows/child/index.htm',
+                    domain: 'mock://www.child.com',
+                    props:  {
+                        account: {
+                            type:       'string',
+                            queryParam: false,
+                            required:   true,
+                            value:      () => undefined
+                        },
+                        testAccountParam: {
+                            type:       'string',
+                            queryParam: 'account',
+                            decorate:   ({ props }) => props.account,
+                            default:    () => '',
+                            required:   false
+                        },
+                        passFoo: {
+                            type:     'function',
+                            required: true
+                        }
+                    }
+                });
+            };
+
+            const component = window.__component__();
+            const instance = component({
+                run: () => `
+                    window.xprops.passFoo({ query: location.search });
+                `,
+                account: '123',
+                passFoo: expect('passFoo', ({ query }) => {
+                    if (query.indexOf(`account=123`) === -1) {
+                        throw new Error(`Expected account=123 in the url, but got ${query}`);
+                    }
+
+                })
+            });
+            
+            return instance.render(getBody());
+        });
+    });
+
     it('should render a component with a prop with a pre-defined value using container, and pass the value in the url', () => {
         return wrapPromise(({ expect }) => {
             const bodyWidth = getBody().offsetWidth;
--- a/test/tests/props.js
+++ b/test/tests/props.js
@@ -41,7 +41,7 @@ describe('zoid props cases', () => {
                     }
                 })
             });
-            
+
             return instance.render(getBody());
         });
     });
@@ -82,7 +82,7 @@ describe('zoid props cases', () => {
                     }
                 })
             });
-            
+
             return instance.render(getBody());
         });
     });
@@ -99,7 +99,9 @@ describe('zoid props cases', () => {
                             type:       'string',
                             queryParam: false,
                             required:   true,
-                            value:      () => undefined
+                            value:      ({ props }) => {
+                                return props.account;
+                            }
                         },
                         testAccountParam: {
                             type:       'string',
@@ -124,12 +126,12 @@ describe('zoid props cases', () => {
                 account: '123',
                 passFoo: expect('passFoo', ({ query }) => {
                     if (query.indexOf(`account=123`) === -1) {
-                        throw new Error(`Expected account=123 in the url, but got ${query}`);
+                        throw new Error(`Expected account=123 in the url, but got ${ query }`);
                     }
 
                 })
             });
-            
+
             return instance.render(getBody());
         });
     });
@@ -171,7 +173,7 @@ describe('zoid props cases', () => {
                     }
                 })
             });
-            
+
             return instance.render(getBody());
         });
     });
@@ -215,7 +217,7 @@ describe('zoid props cases', () => {
             });
 
             Object.defineProperty(document, 'readyState', { value: 'loading', configurable: true });
-            
+
             const renderPromise = instance.render('#container-element');
 
             const container = document.createElement('div');
@@ -257,7 +259,7 @@ describe('zoid props cases', () => {
                     });
                 }
             });
-            
+
             return instance.render(getBody());
         });
     });
@@ -293,7 +295,7 @@ describe('zoid props cases', () => {
                     });
                 }
             });
-            
+
             return instance.render(getBody());
         });
     });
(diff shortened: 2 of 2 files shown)


westeezy:
Opening a Draft PR to ensure we caught the proper issue as I'm not intimately familiar with zoid.

src/parent/props.js
```diff
@@ -11,7 +11,7 @@ import type { ParentHelpers } from './index';
 export function extendProps<P, X>(propsDef : PropsDefinitionType<P, X>, existingProps : PropsType<P>, inputProps : PropsInputType<P>, helpers : ParentHelpers<P>, container : HTMLElement | void) {
     const { state, close, focus, event, onError } = helpers;
 
-    const newProps = { ...existingProps };
+    const newProps = { ...existingProps, ...inputProps };
```
westeezy, line 14: This code use to `extend(props, inputProps)` where `props` has been renamed to `currentProps`. The more I compare the existing to new the less I think this is the correct fix. I'll need to dig a bit more in the morning, but the test case and root idea is hopefully a good start for finding the right fix here.

test/tests/props.js
```diff
@@ -87,6 +87,53 @@ describe('zoid props cases', () => {
...
+                return zoid.create({
+                    tag:    'test-prop-value-container-url-param-bug',
+                    url:    'mock://www.child.com/base/test/windows/child/index.htm',
+                    domain: 'mock://www.child.com',
+                    props:  {
+                        account: {
+                            type:       'string',
+                            queryParam: false,
+                            required:   true,
+                            value:      () => undefined
```
bluepnume, line 102: This test case is strange to me. The intention is that you should either:

a) Pass `value` when defining the prop
b) Pass the value when instantiating the component

...never both at the same time.

bluepnume: Just tested, if I change this to:

```javascript
required: false,
default: () => undefined
```

-- then the test passes as expected.

westeezy: This is the scenario that was unaccounted for in the bug

bluepnume: > This is the scenario that was unaccounted for in the bug

Yeah, so it looks like messaging-components is passing:

```javascript
required: true,
value: () => undefined
```

This is not correct, you can't have a "required but undefined" prop.

### Correct

```javascript
required: true,
value: () => 'foobar'
```

```javascript
required: false,
default: () => 'foobar'
```

```javascript
required: false,
default: () => undefined
```

### Incorrect

```javascript
required: true,
value: () => undefined
```

```javascript
required: true,
default: () => 'foobar'
```

```javascript
required: true,
default: () => undefined
```

bluepnume: So if the props in messaging-components are potentially not defined, the correct way to specify that is to pass:

```javascript
required: false,
default: () => whatever
```

`required: false` -> The value does not need to be passed

`default: () => whatever` -> If there is a viable default it will be used.

westeezy: Actually I think what might be happening now that I see more about how `inputProps` and the `value` function is that on:

```
prop: {
  foo: { 
    value: ({props}) => prop.foo // Is this props param meant to contain the `inputProps`?
  }
}
```

bluepnume: >  Is this props param meant to contain the `inputProps`

Yeah, I think that's the real bug in zoid here. Good catch.

westeezy: Okay I think I have a better idea of what's going on and what I'm trying to say. This mix up is definitely some confusion I had around the value function and the input prop behaviors. It looks like in the value function of a propDef prior to the last zoid release would give a `props` key that contained the values from `inputProps`, but after the latest release it only contains `existingProps`.

For messaging this means that `account` from `inputProps` is not present in the following's props parameter [code](https://github.com/paypal/paypal-messaging-components/blob/f0cd8d1e9b6dce5ad4602ad559e3f4f4824f10d5/src/zoid/message/validation.js#L55).

So I think the "bug" is there and hopefully the updated test shows this a little better. Running locally I can confirm this change in behavior, however if zoid was not intended to behave this way then that might be a different story. 

Regardless I'm basically learning zoid as I dive through this so bear with me

test/tests/props.js
```diff
@@ -82,7 +82,56 @@ describe('zoid props cases', () => {
...
+                    tag:    'test-prop-value-container-url-param-bug',
+                    url:    'mock://www.child.com/base/test/windows/child/index.htm',
+                    domain: 'mock://www.child.com',
+                    props:  {
+                        account: {
+                            type:       'string',
+                            queryParam: false,
+                            required:   true,
+                            value:      ({ props }) => {
+                                return props.account;
```
westeezy, line 103: in here `props.account` was undefined. It's kind of a roundabout test, so I think we can clean it up to not test off this later going into a query param. I just wanted to make the test simulate the bug for the short term so if we do choose to merge this I'll update this to be more concise and clear.

test/tests/props.js
bluepnume: Actually I'm still a little confused here. I'm writing a few quick test cases to verify this behavior and it's working as expected for me. Hold fire, will share a branch

test/tests/props.js
bluepnume: Having `props.account.value` return `props.account` is kinda weird. `value` is only intended for when the caller will not be passing in the value.

westeezy: Yeah I agree with it being confusing as well

bluepnume: Ah, I see this is exactly what messaging-components is doing. Erk.

westeezy: yah! i think that might resolve the confusion on both our ends haha. I'm trying to make something goofy happening and you are seeing its goofy. Granted I think I described it poorly at first

bluepnume: Yeah, ok, the more I look at it, this fix works for me.

Honestly, I kinda hate `extendProps`; it suffers from:

- Recycling the same mutable props object and extending it over and over
- It's never clear when you get `props` in one of these helper functions whether it's only partially constructed, or the default values are populated, or it's decorated, or not.

But I don't think there's much to be done about that right now. We should re-think/overhaul it for the next major version.

In the mean time `{ ...existingProps, ...inputProps }` sounds good.

Thanks for taking the time to investigate!

westeezy: Okay cool I think theres still a few flow errors and such to completely resolve and like I mentioned, I want to make the test read more clearly. I'll tackle those in the morning. 

Yeah super happy to chat more about an overhaul in the future. I'll need to dive deeper into zoid myself to provide more value on that conversation.

I'll mark the PR as "not draft" once its all cleaned up. If you think of anything in the meantime of course let me know and I'm happy to change direction! Thanks for chatting through this, and sorry for the super muddy description I initially provided!

test/tests/props.js
```diff
@@ -82,7 +82,56 @@ describe('zoid props cases', () => {
...
+                            }
+                        },
+                        testAccountParam: {
```
gregjopa, line 106: I like how messaging-components describes this pattern as computed props ([link](https://github.com/paypal/paypal-messaging-components/blob/f0cd8d1e9b6dce5ad4602ad559e3f4f4824f10d5/src/zoid/message/component.js#L313)). Can we add a comment here that describes this behavior? Ex:

```js
    // computed property that depends on the `account` prop
   testAccountParam: {
```

test/tests/props.js
```diff
@@ -82,7 +82,56 @@ describe('zoid props cases', () => {
...
+                            type:       'string',
+                            queryParam: 'account',
```
gregjopa, line 108: This query param being named `account` threw me off. Could we rename it to match the prop name? Or perhaps use `account` and `payerID` for the naming of these two props?
```suggestion
                            queryParam: 'testAccountParam',
```

westeezy: yeah absolutely. We likely don't even need to test this via a query param I just simulated the exact prop defs from messaging, however if we stick with a query param in the test will rename! Let me take a peek at the test today and how I can make it more concise/clear real quick before taking this change

test/tests/props.js
westeezy: Yeah for sure. Just to be clear are you thinking of adding information about that pattern to the docs or do you mean just some comments in the test to describe it? I'm happy to both if we think thats the right idea

westeezy:
We are reverting this release in #378. We will be remerging the test portion of this PR after the revert and then looking into the container prop in the future

PR #378, opened 2021-11-02
Revert "Pass container to prop definitions when available (#376)"

gregjopa:
This reverts commit 89342cc237cd865a73285e5e423ec2e44c2ee437. We found an issue with this `container` prop change and will revert for now.


westeezy (approved):
lgtm

PR #379, opened 2021-11-02
chore: add tests related to zoid prop usage in the wild

westeezy:


test/tests/props.js
```diff
@@ -126,7 +126,87 @@ describe('zoid props cases', () => {
...
+                    }
+                });
+            };
+
+            const component = window.__component__();
+            const instance = component({
+                run: () => `
+                    window.xprops.onSuccess(null, {success: true});
+                `,
+                onSuccess:   expect('onSuccess'),
```
bluepnume, line 206: Can we add an assertion here when `onSuccess` is called that we got the correct argument with a value of `true`?

westeezy: you got it!

wsbrunson (approved):

bluepnume (approved):

gregjopa (approved):

test/tests/props.js
bluepnume: As an academic question, I wonder why Flow allowed this in the first place? Probably the answer is Flow generics not being great. Moving to TS will probably flag up a lot of weird stuff like this where the signature changes.

PR #381, opened 2021-11-11
Pass container to prop definitions when available (round 2)

borodovisin:
### Description
📖  This PR contains the latest commit merged with the @bluepnume commit: 89342cc237cd865a73285e5e423ec2e44c2ee437.
This PR contains a bug fix related to the decorate functions not working properly due to the override reference after the container parameter was introduced.

👀  The problem was related to this function `setProps`. Now we're calling that function twice in the zoid code. One in the instance creation step and two in the render process. That causes a reference function overriding in this file `parent/props.js` in the `extendProps` utils function.

If we call twice this`propDef.decorate` function, that will return a `new` function, meaning the second time we call the `propDef.decorate` won't be the original defined function, instead will have a reference to the result of the execution of the decorated function.

### Why are we making these changes? 
We're making this change to solve this failing test case: `should allow decorated functions to change their signature from the original value callback prop`. After the new `container` parameter was added to the decorator this test starts failing.

### Reproduction Steps
If we set any kind of decorate logic that uses the `value` argument, the code won't work. 
You can use this logic from the failing test case:

🧪  
```
 decorate: ({ value, onError }) => {
    return (err, result) => {
        if (err || !result) {
            return onError(err);
        }

        return value(true);
    };
}

 window.xprops.onSuccess(null, {success: true});
```
:boom: The first time the function is called, the `err` argument will be equal to: `null` and `result` argument equal to: `{success: true}`. In the second call, the `err` argument will be equal to: `true` and `result` argument equal to: `undefined`.

🤔 Could be a good idea to use `Proxy`objects, or maybe `Object.setPrototypeOf` to have a better control over what we're modifying in the `propDef`. Both of them aren't support in Internet Explorer 💣

### DIFF
--- a/src/component/component.js
+++ b/src/component/component.js
@@ -4,14 +4,14 @@
 import { setup as setupPostRobot, on, send, bridge, toProxyWindow, destroy as destroyPostRobot } from 'post-robot/src';
 import { ZalgoPromise } from 'zalgo-promise/src';
 import { isWindow, getDomain, matchDomain, type CrossDomainWindowType, type DomainMatcher } from 'cross-domain-utils/src';
-import { noop, isElement, cleanup, memoize, identity, extend, uniqueID } from 'belter/src';
+import { noop, isElement, cleanup, memoize, identity, extend, uniqueID, values } from 'belter/src';
 
 import { childComponent, type ChildComponent } from '../child';
 import { type RenderOptionsType, type ParentHelpers, parentComponent } from '../parent/parent';
 import { ZOID, CONTEXT, POST_MESSAGE, WILDCARD, METHOD, PROP_TYPE } from '../constants';
 import { react, angular, vue, vue3, angular2 } from '../drivers';
 import { getGlobal, destroyGlobal, getInitialParentPayload, isChildComponentWindow } from '../lib';
-import type { CssDimensionsType, StringMatcherType } from '../types';
+import type { CssDimensionsType, StringMatcherType, ContainerReferenceType } from '../types';
 
 import { validateOptions } from './validate';
 import { defaultContainerTemplate, defaultPrerenderTemplate } from './templates';
@@ -135,8 +135,8 @@ export type ZoidComponentInstance<P, X = void, C = void> = {|
     ...C,
     isEligible : () => boolean,
     clone : () => ZoidComponentInstance<P, X, C>,
-    render : (container? : string | HTMLElement, context? : $Values<typeof CONTEXT>) => ZalgoPromise<void>,
-    renderTo : (target : CrossDomainWindowType, container? : string | HTMLElement, context? : $Values<typeof CONTEXT>) => ZalgoPromise<void>
+    render : (container? : ContainerReferenceType, context? : $Values<typeof CONTEXT>) => ZalgoPromise<void>,
+    renderTo : (target : CrossDomainWindowType, container? : ContainerReferenceType, context? : $Values<typeof CONTEXT>) => ZalgoPromise<void>
 |};
 
 // eslint-disable-next-line flowtype/require-exact-type
@@ -270,7 +270,6 @@ export function component<P, X, C>(opts : ComponentOptionsType<P, X, C>) : Compo
     if (__DEBUG__) {
         validateOptions(opts);
     }
-
     const options = normalizeOptions(opts);
 
     const {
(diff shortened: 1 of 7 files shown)


src/component/component.js
```diff
@@ -254,6 +254,30 @@ function normalizeOptions<P, X, C>(options : ComponentOptionsType<P, X, C>) : No
...
 
+/**
+ * Modify the decorate properties definition function to run once only.
+ * Currently we have a problem about overriding function reference if we called
+ * setProps twice, and now that is the case with the new container argument.
+ * 
+ * @param opts the defined options in the component creation step
+ * @return the same opts object with decorate function modified to run once if exists
+ */
+function decorateRunOnceModifier(opts : ComponentOptionsType<P, X, C>): ComponentOptionsType<P, X, C> {
```
gregjopa, line 265: This approach seems reasonable to me! Two things come to mind:
1. Let's use `Object.keys()` instead of `Object.values()` and `Object.entries` since its safe to use with the current browser support requirements.
2. Let's flatten out this code to improve readability. Also, I'm wondering if we can call `decorateRunOnceModifier()` inside of `normalizeOptions` so it's all in one spot.

Example of flattening out the code:
```js
function decorateRunOnceModifier(opts : ComponentOptionsType<P, X, C>): ComponentOptionsType<P, X, C> {
    if (!isObject(opts) {
        return opts;
    }

    if (!isObject(opts.props)) {
        return opts;
    }

    Object.values(opts.props).forEach(parentValue => {
        if (!isObject(parentValue)) {
            return;
        }

        Object.entries(parentValue).forEach(([key, value]) => {
            if (key === "decorate") {
                parentValue[key] = once(value);
            }
        });
    });


    return opts;

}

```

borodovisin: This solution won't apply

gregjopa:
@bluepnume and @westeezy please review when you get a chance. I'm hoping we can release a beta version of zoid next week to test out this change with other components.

src/parent/props.js
```diff
@@ -1,103 +1,91 @@
...
-        const propDef = propsDef[key];
-
-        // $FlowFixMe
-        let value = inputProps[key];
+    
+    /* NOTE: The inputProps (PropsInputType) type is not compatible with
+        the existingProps (PropsType) type,
+        that is the reason to removed the parent key from that object and
+        merge with the existingProps object
+        eslint-disable-next-line no-unused-vars
```
westeezy, line 18: Can you remove the double `eslint-disable-next-line`?

bluepnume:
Apologies for the delay! Taking a look as we speak.

src/parent/props.js
```diff
@@ -1,103 +1,90 @@
...
 
         if (isDefined(value) && propDef.decorate) {
-            // $FlowFixMe[incompatible-call]
-            const decoratedValue = propDef.decorate({ value, props, state, close, focus, event, onError });
-            // $FlowFixMe[incompatible-type]
-            props[key] = decoratedValue;
+            // Store the value the first time the decorate props is executed and use it in further executions
+            // This will be delete if the component instance is cloned
+            // $FlowFixMe
+            const originalValue = getOrSet(propDef, '_originalDefinition', () => value);
```
bluepnume, line 76: Yeah this is going to be really problematic -- since the same `propDef` object could be shared between many different instances of a component that are instantiated/rendered at the same time.

Lemme think through if there's a better alternative, but if we want to store the original value I think we should find a scope that's local to the current component instance.

src/component/component.js
```diff
@@ -414,6 +413,13 @@ export function component<P, X, C>(opts : ComponentOptionsType<P, X, C>) : Compo
         });
 
         const clone = ({ decorate = identity } = {}) => {
+            // Delete any existence of the decorate function before clone the instance
+            // The _originalDefinition was added in the parent/props.js in the decorate logic
+            // $FlowFixMe
+            values(propsDef).forEach(prop => {
+                // $FlowFixMe
+                delete prop._originalDefinition;
```
bluepnume, line 421: Not a big fan of making `clone` a special case where we do cleanup. There are plenty of other cases where the same component can be rendered multiple times. Also clone does not imply that the original component is not needed any more.

src/parent/props.js
bluepnume: FYI: added a few tests which cover multiple component instances, and multiple `updateProps` calls, please see https://github.com/krakenjs/zoid/commit/044028fcfb199707bf388715f55f1871aef0c23e

bluepnume: OK, pushed up a new branch/commit which avoids the multi-decorate pattern and also passes all of these new tests. Please let me know if this looks ok to you: https://github.com/krakenjs/zoid/commit/c4bba5c1638cca17b618b99abf01cf3804520df4

borodovisin: Looks like a cleaner code instead of storing all the time a function reference.

borodovisin:
Hi, @bluepnume thanks for the review. The solution keeping a copy of the reference isn't a good solution. Go ahead with your PR, I'll decline this one.

borodovisin:
@bluepnume make a better solution related to the decorate function called twice in component creation and render process.

We do not need to iter over all the `newProps` just over the inputProps.

PR #390, opened 2022-02-21
Update Github Actions

westeezy:
- Add CodeCov
- Make consistent filename and scripts
- Remove legacy travis

Note: eventually we should decouple build and test but having everything
consistent is better than inconsistent to me.


gregjopa (approved):
💯

Issue #393, opened 2022-02-22
driver.register does not pass down context to frameworks


artmanque:
As per the docs:
[defaultContext](https://github.com/krakenjs/zoid/blob/main/docs/api/create.md#defaultcontext-string)
[opts.context](https://github.com/krakenjs/zoid/blob/main/docs/api/create.md#opts)
arguments to [zoid.create](https://github.com/krakenjs/zoid/blob/main/docs/api/create.md#create-new-component) are not passed down into framework's register method. All the frameworks use hard constant `CONTEXT.IFRAME` overriding create-component arguments

### Steps to Reproduce:
Live Sandbox: https://codesandbox.io/s/intelligent-wave-8nsb55

PR #394, opened 2022-02-22
Fix Frameworks defaultContext option

artmanque:
`defaultContext` option from `zoid.create` is not used inside framework's `register` method to pass it on to `render` method of `ZoidComponentInstance`


artmanque:
fixes #393

PR #401, opened 2022-04-20
chore: use prettier

westeezy:


bluepnume:
Can we start with a prettier config that approximately matches the current lint style settings? (e.g. single quotes for strings)

westeezy:
The goal overall to move style rules out of linting and use a base prettier config. So the first question I have: is there a concern you have with moving to prettier's default config? I know it will be a large diff but hoping back commits is quite easy in most editors. Also if we use prettier we will have to forgo at least some style rules even if we attempted to match such as the spacing of object keys, so if the goal is to keep the same style rules we have now then I would say no to prettier and keep them in eslint.

bluepnume:
Sure; I mean at least with the single quote thing:

- It's a minor tweak in the config that saves on a huge number of unnecessary changes
- Single quotes are easier to type. Can type them one handed, with no shift key. 😂

westeezy:
> Sure; I mean at least with the single quote thing:
> 
> * It's a minor tweak in the config that saves on a huge number of unnecessary changes
> * Single quotes are easier to type. Can type them one handed, with no shift key. 😂

Ah perfect I just want to be sure we are on the same page in terms of moving to prettier is agreed upon. So in the spirit of changes to prettier I think it'll likely be single quotes, 4 units on indent, and 120 line length. I'd have to check if anything else in the docs quickly. I'll also have to double check if we were tabs or spaces quickly since I just set my editors to auto-format to our rules in grumbler-scripts and didn't pay attention.

bluepnume:
Ya for sure, totally happy with prettier, just prefer to do things incrementally when possible. So yeah would be good to stick with single quotes / 4 spaces / etc. for now.

wsbrunson (approved):

westeezy:
After chatting with the larger team a bit more we have concluded that we would like to pursue the base prettier config and that its probably easiest to just rip the bandaid all at once so we are going to merge this with base config. Of course that means git blaming will typically require you to go back an extra commit but eventually that'd have to be the case to get to the base config for prettier anyways.

Issue #334, opened 2021-01-06
Removing the react container component after close() or destroy() results in error  


tmilar:
Hi, I've running into this trouble and I couldn't sort it out.

I'm using React v17.

* I'm opening an HTML modal, which instantiates and renders a Zoid component, as an _iframe_.
* Click on the background to close it, which runs: `zoidComponent.close()` to clean it up.

If enough time has passed in between, so the iframe loaded and rendered properly, this results in no errors ✅ 

**But**  if I close it fast enough, then I get this error 🛑:

```
zoid.frameworks.frame.js:3201 Uncaught Error: Detected container element removed from DOM
    at zoid.frameworks.frame.js:3201
    at anonymous::once (zoid.frameworks.frame.js:1045)
    at elementClosed (zoid.frameworks.frame.js:3165)
    at removeChild (react-dom.development.js:10301)
    at unmountHostComponents (react-dom.development.js:21296)
    at commitDeletion (react-dom.development.js:21347)
    at commitMutationEffects (react-dom.development.js:23407)
    at HTMLUnknownElement.callCallback (react-dom.development.js:3945)
    at Object.invokeGuardedCallbackDev (react-dom.development.js:3994)
    at invokeGuardedCallback (react-dom.development.js:4056)
    at commitRootImpl (react-dom.development.js:23121)
    at unstable_runWithPriority (scheduler.development.js:646)
    at runWithPriority$1 (react-dom.development.js:11276)
    at commitRoot (react-dom.development.js:22990)
    at performSyncWorkOnRoot (react-dom.development.js:22329)
    at react-dom.development.js:11327
    at unstable_runWithPriority (scheduler.development.js:646)
    at runWithPriority$1 (react-dom.development.js:11276)
    at flushSyncCallbackQueueImpl (react-dom.development.js:11322)
    at flushSyncCallbackQueue (react-dom.development.js:11309)
    at scheduleUpdateOnFiber (react-dom.development.js:21893)
    at dispatchAction (react-dom.development.js:16139)
    at Object.onDestroy (App.tsx:61)
    at zoid.frameworks.frame.js:3320
    at zoid.frameworks.frame.js:2746
    at Function.ZalgoPromise.try (zoid.frameworks.frame.js:387)
    at _loop (zoid.frameworks.frame.js:2745)
    at Object.trigger (zoid.frameworks.frame.js:2749)
    at zoid.frameworks.frame.js:2967
[...]

PR #424, opened 2023-04-07
fix: avoid throwing errors if container is no longer in page

oscarleonnogales:
### Description

We have identified a few errors that zoid throws that cannot be caught by the implementation code (`render.catch() `and `close.catch()`) during asynchronous dismounts. Although this can occur in every webpage, it's primarily an issue when using React as seen here: [Issue #334](https://github.com/krakenjs/zoid/issues/334)

With this change, we first check if the container element has been removed from the DOM or the user has navigated away from the page mid-render. In these scenarios we should avoid throwing errors.

Here are 3 examples:
https://github.com/krakenjs/zoid/blob/main/src/parent/parent.js#L857
https://github.com/krakenjs/zoid/blob/main/src/parent/parent.js#L826
https://github.com/krakenjs/zoid/blob/main/src/parent/parent.js#L1172

These 3 error flows eventually cause the promise to be rejected since the render was never completed, and an error to be displayed on the console: https://github.com/krakenjs/zoid/blob/main/src/parent/parent.js#L732

### Reproduction Steps (if applicable)

To reproduce the `Window navigated away` error, open Chrome developer tools and persist logs. Call `paypalButtons.render('container-selector')` and then refresh the window or navigate to a different URL before second render is finished.

To reproduce `Component closed` or `Detected container element removed from DOM`, Call `paypalButtons.render('container-selector')` and then immediately remove the container selector before second render is finished.

### Why are we making these changes? Include references to any related Jira tasks or GitHub Issues

https://paypal.atlassian.net/browse/LI-7432

### DIFF
--- a/src/constants.js
+++ b/src/constants.js
@@ -9,7 +9,7 @@ export const POST_MESSAGE = {
   ALLOW_DELEGATE: `${ZOID}_allow_delegate`,
 };
 
-export const RENDER_ERRORS = {
+export const COMPONENT_ERROR = {
   NAVIGATED_AWAY: "Window navigated away",
   COMPONENT_DESTROYED: "Component destroyed",
   COMPONENT_CLOSED: "Component closed",
--- a/src/constants.js
+++ b/src/constants.js
@@ -9,6 +9,13 @@ export const POST_MESSAGE = {
   ALLOW_DELEGATE: `${ZOID}_allow_delegate`,
 };
 
+export const RENDER_ERRORS = {
+  NAVIGATED_AWAY: "Window navigated away",
+  COMPONENT_DESTROYED: "Component destroyed",
+  COMPONENT_CLOSED: "Component closed",
+  WINDOW_CLOSED: "Detected component window close",
+};
+
 export const PROP_TYPE = {
   STRING: ("string": "string"),
   OBJECT: ("object": "object"),
--- a/src/constants.js
+++ b/src/constants.js
@@ -14,6 +14,8 @@ export const COMPONENT_ERROR = {
   COMPONENT_DESTROYED: "Component destroyed",
   COMPONENT_CLOSED: "Component closed",
   WINDOW_CLOSED: "Detected component window close",
+  POPUP_CLOSE: "Detected popup close",
+  IFRAME_CLOSE: "Detected iframe close",
 };
 
 export const PROP_TYPE = {
--- a/src/parent/parent.js
+++ b/src/parent/parent.js
@@ -317,7 +317,6 @@ export function parentComponent<P, X, C>({
   let childComponent: ?ChildExportsType<P>;
   let currentChildDomain: ?string;
   let currentContainer: HTMLElement | void;
-  let isRenderFinished: boolean = false;
 
   const onErrorOverride: ?OnError = overrides.onError;
   let getProxyContainerOverride: ?GetProxyContainer =
@@ -1619,7 +1618,6 @@ export function parentComponent<P, X, C>({
       });
 
       const onRenderedPromise = initPromise.then(() => {
-        isRenderFinished = true;
         return event.trigger(EVENT.RENDERED);
       });
 
--- a/src/parent/parent.js
+++ b/src/parent/parent.js
@@ -316,7 +316,7 @@ export function parentComponent<P, X, C>({
   let currentProxyContainer: ?ProxyObject<HTMLElement>;
   let childComponent: ?ChildExportsType<P>;
   let currentChildDomain: ?string;
-  let currentContainer: HTMLElement | void;
+  let currentContainer: HTMLElement;
 
   const onErrorOverride: ?OnError = overrides.onError;
   let getProxyContainerOverride: ?GetProxyContainer =
--- a/src/parent/parent.js
+++ b/src/parent/parent.js
@@ -65,7 +65,7 @@ import {
   METHOD,
   WINDOW_REFERENCE,
   DEFAULT_DIMENSIONS,
-  RENDER_ERRORS,
+  COMPONENT_ERROR,
 } from "../constants";
 import {
   getGlobal,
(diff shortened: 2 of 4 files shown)


wsbrunson (commented):
I haven't looked into testing in Zoid so I don't know how reasonable this ask is: can we write a test for this?

src/parent/parent.js
```diff
@@ -729,7 +729,16 @@ export function parentComponent<P, X, C>({
         return clean.all(err);
       })
       .then(() => {
-        initPromise.asyncReject(err || new Error("Component destroyed"));
+        const error = err || new Error("Component destroyed");
+        if (
+          isElementClosed(currentContainer) ||
+          error.message === "Window navigated away"
+        ) {
+          console.warn(error);
```
gregjopa, line 737: I'm leaning towards not logging anything to the console for this use case. If the window is closed or the user navigated away from the page then we can have destroy be silent. 
```suggestion
```

src/parent/parent.js
```diff
@@ -729,7 +729,16 @@ export function parentComponent<P, X, C>({
         return clean.all(err);
       })
       .then(() => {
-        initPromise.asyncReject(err || new Error("Component destroyed"));
+        const error = err || new Error("Component destroyed");
+        if (
+          isElementClosed(currentContainer) ||
+          error.message === "Window navigated away"
```
gregjopa, line 735: Can we create a top level variable for this error message? Seems like it's coming from here: https://github.com/krakenjs/zoid/blob/main/src/parent/parent.js#L826. It would be nice if both of these code paths shared the variable for this error message so if it changes in the future, this wouldn't break.

gregjopa:
> I haven't looked into testing in Zoid so I don't know how reasonable this ask is: can we write a test for this?

I agree that it would be awesome to add a couple tests to cover this new logic. I'm really excited about this improvement. Great work so far @oscarleonnogales!

src/parent/parent.js
```diff
@@ -315,7 +316,7 @@ export function parentComponent<P, X, C>({
   let currentProxyContainer: ?ProxyObject<HTMLElement>;
   let childComponent: ?ChildExportsType<P>;
   let currentChildDomain: ?string;
-  let currentContainer: HTMLElement | void;
```
gregjopa, line 318: I'm a little concerned about removing the `| void` here. Wondering if we can add it back and instead add a check to make sure it's truthy before passing it to `isElementClosed()`. Ex:

```js
        const error = err || new Error(RENDER_ERRORS.COMPONENT_DESTROYED);
        if (
          (currentContainer && isElementClosed(currentContainer)) ||
          error.message === RENDER_ERRORS.NAVIGATED_AWAY
        ) {
          initPromise.resolve();
        } else {
          initPromise.asyncReject(error);
        }
```

src/constants.js
```diff
@@ -9,6 +9,13 @@ export const POST_MESSAGE = {
   ALLOW_DELEGATE: `${ZOID}_allow_delegate`,
 };
 
+export const RENDER_ERRORS = {
```
gregjopa, line 12: Great to see constants for these error messages 💯

gregjopa, line 12: Nit: all the other constants are singular in this file. It would be nice to follow that same naming convention.
```suggestion
export const RENDER_ERROR = {
```

mdentremont:
This is great, thanks everyone!

src/constants.js
gregjopa: Also, what do you think about the name `COMPONENT_ERROR` instead of `RENDER_ERROR`. I think some of these errors can occur after the render phase is complete.

gregjopa (approved):
This looks great to me! Nice work @oscarleonnogales 💯

spencersablan:
Looks great man 🔥  Love that you moved the error messages into a const ⚡️

spencersablan (approved):

src/parent/parent.js
```diff
@@ -853,8 +863,15 @@ export function parentComponent<P, X, C>({
...
         if (!cancelled) {
-          if (isClosed) {
-            return close(new Error(`Detected ${context} close`));
+          if (context === CONTEXT.POPUP && isClosed) {
+            return close(new Error(COMPONENT_ERROR.POPUP_CLOSE));
+          } else if (
+            context === CONTEXT.IFRAME &&
+            isClosed &&
+            ((currentContainer && isElementClosed(currentContainer)) ||
+              isSecondRenderFinished)
```
wsbrunson, line 872: where did we land on checking for second render? I thought we said this might cause weird issues but I don't remember if we changed our minds later

oscarleonnogales: We did decide to not use the condition of waiting until second render was finished by itself, as that could cause issues when the container is removed for a legitimate reason before it's finished.

Essentially I combined options 1 and 2 and it seemed to work well. If the container is removed, OR second render is finished then we should call `close` here.

wsbrunson: okay sounds good, thank you!

wsbrunson (approved):

src/parent/parent.js
gregjopa: I am a little worried about adding the new isSecondRenderFinished logic here. The prerender phase still renders inside an `<iframe>` with no src tag. If that gets destroyed, we would still want this error to be thrown.

src/parent/parent.js
```diff
@@ -316,6 +317,7 @@ export function parentComponent<P, X, C>({
   let childComponent: ?ChildExportsType<P>;
   let currentChildDomain: ?string;
   let currentContainer: HTMLElement | void;
+  let isSecondRenderFinished: boolean = false;
```
gregjopa, line 320: Can we rename this to render? I think the existing naming convention in zoid is prerender (phase 1) and render (phase 2) 
```suggestion
  let isRenderFinished: boolean = false;
```

src/parent/parent.js
```diff
@@ -1609,6 +1619,7 @@ export function parentComponent<P, X, C>({
       });
 
       const onRenderedPromise = initPromise.then(() => {
+        isRenderFinished = true;
```
gregjopa, line 1622: Could we remove this variable from this PR and move it to the next PR?
```suggestion
```

src/parent/parent.js
```diff
@@ -316,6 +317,7 @@ export function parentComponent<P, X, C>({
   let childComponent: ?ChildExportsType<P>;
   let currentChildDomain: ?string;
   let currentContainer: HTMLElement | void;
+  let isRenderFinished: boolean = false;
```
gregjopa, line 320:
```suggestion
```

PR #425, opened 2023-04-24
fix: only throwing errors for major version changes

oscarleonnogales:
### Description

We currently throw an error when the Zoid version doesn't strictly match. This change will only check for major version upgrades before throwing an error, patches and minor changes will be allowed to be different.

### DIFF
--- a/src/child/child.js
+++ b/src/child/child.js
@@ -74,6 +74,10 @@ function destroy(): ZalgoPromise<void> {
   });
 }
 
+function hasMajorVersionChanged(version1: string, version2: string): boolean {
+  return version1.split("_")[0] !== version2.split("_")[0];
+}
+
 export type ChildComponent<P, X> = {|
   getProps: () => ChildPropsType<P, X>,
   init: () => ZalgoPromise<void>,
@@ -100,7 +104,7 @@ export function childComponent<P, X, C>(
     props: initialProps,
   } = payload;
 
-  if (version !== __ZOID__.__VERSION__) {
+  if (hasMajorVersionChanged(version, __ZOID__.__VERSION__)) {
     throw new Error(
       `Parent window has zoid version ${version}, child window has version ${__ZOID__.__VERSION__}`
     );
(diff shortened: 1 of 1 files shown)


src/child/child.js
```diff
@@ -74,6 +74,15 @@ function destroy(): ZalgoPromise<void> {
...
 }
 
+// zoid version strings should be in snake_case format (10_1_0, 10_2_0, 11_0_0, etc.)
+// since it is saved on the window object as a key. This function will compare the first numerical value
+// that should only change during major changes.
+function hasMajorVersionChanged(version1: string, version2: string): boolean {
+  if (/_/.test(version1) && /_/.test(version2)) {
+    return version1.split("_")[0] !== version2.split("_")[0];
+  }
+}
```
gregjopa, line 84: What do you think about updating this function to always return a boolean? Ex:

```ts
function hasMajorVersionChanged(version1: string, version2: string): boolean {
  // validate that the version numbers are in the expected format
  if (!/_/.test(version1) || !/_/.test(version2) {
    return false;
  }

  return version1.split("_")[0] !== version2.split("_")[0];
}
```

wsbrunson (approved):

spencersablan (approved):

src/child/child.js
```diff
@@ -74,6 +74,22 @@ function destroy(): ZalgoPromise<void> {
   });
 }
 
+// This function will only compare the first numerical value,
```
gregjopa, line 77:
```suggestion
// Compares the first numerical value of the parent and child versions of zoid
// to help throw an error when the major version number is different.
```

src/child/child.js
```diff
@@ -74,6 +74,22 @@ function destroy(): ZalgoPromise<void> {
   });
 }
 
+// Compares the first numerical value of the parent and child versions of zoid,
+// ensuring that an error is only thrown when major versions of Zoid are released.
```
gregjopa, line 78:
```suggestion
// ensuring that an error is only thrown when major versions of Zoid do not match.
```

PR #426, opened 2023-04-26
Reduce occurrence of "Detected iframe close" error

oscarleonnogales:
### Description

We're seeing rare cases when we see `Detected iframe close` before second render has a chance to finish in slow browsers with poor connection. With this change, we're adding an additional check before we decide to throw the error. 

Here is the example:
https://github.com/krakenjs/zoid/blob/main/src/parent/parent.js#L866

### Why are we making these changes? Include references to any related Jira tasks or GitHub Issues

https://paypal.atlassian.net/browse/LI-7432

### DIFF
--- a/src/parent/parent.js
+++ b/src/parent/parent.js
@@ -316,7 +316,7 @@ export function parentComponent<P, X, C>({
   let currentProxyContainer: ?ProxyObject<HTMLElement>;
   let childComponent: ?ChildExportsType<P>;
   let currentChildDomain: ?string;
-  let currentContainer: HTMLElement | void;
+  let currentContainer: HTMLElement;
 
   const onErrorOverride: ?OnError = overrides.onError;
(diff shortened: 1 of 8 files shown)


src/parent/parent.js
```diff
@@ -862,8 +863,18 @@ export function parentComponent<P, X, C>({
       })
       .then((isClosed) => {
         if (!cancelled) {
-          if (isClosed) {
-            return close(new Error(`Detected ${context} close`));
+          const isCurrentContainerClosed: boolean =
+            // $FlowIssue
+            currentContainer && isElementClosed(currentContainer);
```
gregjopa, line 868: Are you able to remove the `$FlowIssue` statement by using Boolean() like so?
```suggestion
            Boolean(currentContainer && isElementClosed(currentContainer));
```

gregjopa (approved):

PR #429, opened 2023-05-04
fix: reduce frequency of iframe removed error

oscarleonnogales:
### Description

We're seeing rare cases when we see `Detected iframe close` before second render has a chance to finish in slow browsers with poor connection. With this change, we're adding an additional check before we decide to throw the error. 

Here is the example:
https://github.com/krakenjs/zoid/blob/main/src/parent/parent.js#L866


gregjopa:
@oscarleonnogales @dtjones404 @westeezy is it cool if we merge this and cut a new release of zoid?

PR #431, opened 2023-05-12
fix: deprecate script driver

dtjones404:


src/drivers/script.js
```diff
@@ -27,22 +27,12 @@ export let script : ComponentDriverType<*, Document> = {
...
-
-            if (!element.parentNode) {
-                throw new Error(`Element has no parent`);
-            }
-
-            element.parentNode.replaceChild(container, element);
-
-            // $FlowFixMe
-            component.render(props, container);
+            throw new Error(`This integration pattern is no longer supported.
```
gregjopa, line 32: My initial thought is we should include this error logic in checkout.js v4 source code, not zoid. 

Could we update checkout.js to do something like 

```js
if ($('script[type="application/x-component"]') {
  // if this special script type is found in the DOM log an error
  console.error(
`This integration pattern is no longer supported.
Please visit https://developer.paypal.com/demo/checkout-v4/#/pattern/client for an example of the new recommended integration pattern.
`);

}
```

gregjopa (approved):

PR #439, opened 2023-06-10
feat(venmo): allow trusted domains

mnicpt:
Allows trusted domains to be specified for passing props to children with different domain.
https://paypal.atlassian.net/browse/DTCHKINT-1568

### DIFF
--- a/src/child/props.js
+++ b/src/child/props.js
@@ -82,11 +82,12 @@ export function normalizeChildProps<P, X>(
   for (const key of Object.keys(props)) {
     const prop = propsDef[key];
 
-    if (
-      prop &&
-      prop.sameDomain &&
-      (origin !== getDomain(window) || !isSameDomain(parentComponentWindow))
-    ) {
+    const trustedChild: boolean =
+      prop && prop.trustedDomains && prop.trustedDomains.length > 0
+        ? prop.trustedDomains.indexOf(getDomain(window)) !== -1
+        : origin === getDomain(window) || isSameDomain(parentComponentWindow);
+
+    if (prop && prop.sameDomain && !trustedChild) {
       continue;
     }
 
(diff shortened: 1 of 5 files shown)


src/child/props.js
```diff
@@ -82,11 +82,12 @@ export function normalizeChildProps<P, X>(
...
     const prop = propsDef[key];
 
-    if (
-      prop &&
-      prop.sameDomain &&
-      (origin !== getDomain(window) || !isSameDomain(parentComponentWindow))
-    ) {
+    const trustedChild: boolean =
+      prop && prop.trustedDomains && prop.trustedDomains.length > 0
+        ? prop.trustedDomains.indexOf(getDomain(window)) !== -1
```
wsbrunson, line 87:
```suggestion
        ? prop.trustedDomains.includes(getDomain(window))
```
we don't really care about IE11 support any more

src/parent/parent.js
```diff
@@ -378,11 +378,12 @@ export function parentComponent<P, X, C>({
...
       }
 
-      if (
-        prop &&
-        prop.sameDomain &&
-        !matchDomain(initialChildDomain, getDomain(window))
-      ) {
+      const trustedChild: boolean =
+        prop && prop.trustedDomains && prop.trustedDomains.length > 0
+          ? prop.trustedDomains.indexOf(initialChildDomain) !== -1
```
wsbrunson, line 383:
```suggestion
          ? prop.trustedDomains.includes(initialChildDomain)
```

test/tests/props.js
```diff
@@ -813,6 +813,73 @@ describe("zoid props cases", () => {
     });
   });
 
+  it("should pass a trustedDomain prop and have it populate on the child", () => {
```
wsbrunson, line 816: we should add a test where the domains don't match 

`url: "mock://www.child.com/base/test/windows/child/index.htm",`
`trustedDomains: ["mock://www.different-domain.com"],`

wsbrunson (approved):

PR #440, opened 2023-06-13
fix(prop): trusted domain

mnicpt:
Changes trusted domain prop to accept regex rather than strings.

### DIFF
--- a/src/parent/parent.js
+++ b/src/parent/parent.js
@@ -380,7 +380,9 @@ export function parentComponent<P, X, C>({
 
       const trustedChild: boolean =
         prop && prop.trustedDomains && prop.trustedDomains.length > 0
-          ? prop.trustedDomains.includes(initialChildDomain)
+          ? prop.trustedDomains.reduce((acc, val) => {
+              return acc || matchDomain(val, initialChildDomain);
+            }, false)
           : matchDomain(initialChildDomain, getDomain(window));
 
       // let trustedDomains override sameDomain prop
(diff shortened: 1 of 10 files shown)


wsbrunson (approved):

src/parent/parent.js
```diff
@@ -380,7 +380,9 @@ export function parentComponent<P, X, C>({
 
       const trustedChild: boolean =
         prop && prop.trustedDomains && prop.trustedDomains.length > 0
-          ? prop.trustedDomains.includes(initialChildDomain)
+          ? prop.trustedDomains.reduce((acc, val) => {
```
gregjopa, line 383: What do you think about accepting a string or regex? We may be able to reuse code from `allowedParentDomains`: https://github.com/search?q=repo%3Akrakenjs%2Fzoid%20allowedParentDomains&type=code

mnicpt: @gregjopa I initially was going down this path until I realized using a string won't play nice when going from environment to environment like Regex does. We could always have it point to an enum. 🤔

mnicpt: The `domain` prop does the same thing as `allowedParentDomain` in that it uses the same type def as `DomainMatcher` from `cross-domain-utils`. I just recently updated `DomainMatcher` to except an array of string or regex.

mnicpt: @gregjopa Yeah, I can do this. I'm using `matchDomain` which allows for all.

gregjopa: >@gregjopa I initially was going down this path until I realized using a string won't play nice when going from environment to environment like Regex does. 

Ahh I see. I'm glad you already considered this.

PR #443, opened 2023-10-30
fix(venmo): window name

mnicpt:
## Issue
Window name wasn't being updated if redirecting in popup. This caused blank windows to open when sending focus event to popup as window name did not match.

## Solution
Send updated name to parent initialization to update the proxy window name.


nbierdeman (approved):
Nice!!! 🚀

PR #464, opened 2024-12-09
feat: support additional properties to be added to zoid component from the client

ravishekhar:
* We have a use case where its necessary to add support for additional methods on the Zoid component instance. 
* To support this use case  `getExtensions` callback is being added to create method.
* Zoid would call this method to add the additional properties returned by this callback on the component instances. 
* This method will accept the parent props as an argument so that the extended methods could interact with props passed during initialization. 
* The implementation will ensure that `getExtensions` cannot override any of the existing methods in Zoid like render, clone etc. 

Sample usage

```javascript
const MyComponent = zoid.create({
    tag: "custom-buttons",
    getExtensions: (parent) => {
     // methods defined in parent props, like `onCancel` will be available in 
     // props returned by parent.getProps();
      return {
        hasReturned: () => true,
        resume: () => {
          console.log("Resuming ....");
        },
      };
    }
});

const component = MyComponent({ 
     onCancel: function() { ... }
})
```
Now the `component` instance will contain the additional methods along with standard methods. 
e.g.
```javascript
component.hasReturned(); // will return true
component.resume(); // will log Resuming .... on console. 
```

### DIFF
--- a/src/component/component.js
+++ b/src/component/component.js
@@ -29,6 +29,7 @@ import {
 
 import { childComponent, type ChildComponent } from "../child";
 import {
+  type ParentComponent,
   type RenderOptionsType,
   type ParentHelpers,
   parentComponent,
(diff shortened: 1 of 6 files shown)


src/component/component.js
```diff
@@ -106,9 +106,10 @@ export type ExportsDefinition<X> =
   | ExportsConfigDefinition
   | ExportsMapperDefinition<X>;
 
-export type ComponentOptionsType<P, X, C> = {|
+export type ComponentOptionsType<P, X, C, ExtType> = {|
   tag: string,
 
+  getExtensions?: (parentProps: X) => ExtType,
```
wsbrunson, line 112: love the name!

wsbrunson (approved):

wsbrunson (approved):

PR #471, opened 2025-05-07
fix: remove deprecated use of findDOMNode

onurio:


sebsobseb (approved):

sebsobseb:
@ravishekhar is there any chance of getting this merged / released any time soon?

joegirgis:
Can we get this merged to support React 19 please? we are still using Zoid and had to revert the React update because it's not working with React 19

Issue #130, opened 2017-11-05
Notify child when a parent prop/value changes


cmackenz:
Struggling a little bit on what seems like an easy concept.

I am passing props down to my child component and that works fine, but I want my child component to re-render when the a parent value changes. 

A trivial example would be something like:

```
  let counter = 0;
  setInterval(() => {
    counter++;
  }, 1000);

  COUNTER_WIDGET.render({
    counter,
  }, '#counter');
```

And then in my child component (Ember)
```
export default Ember.Route.extend({

  setupController(controller, model) {
    this._super(...arguments);
    set(controller, 'attrs', {
      counter: window.xprops.counter,
    })

    setInterval(() => {
      console.log('window.xprops.counter', window.xprops.counter);
      // Always 0
    }, 1000)
  },

})
```

How do I get the new value of `counter` in my child component when it has been updated on the parent?

freshfunkee:
Not sure if you already received an answer, but a child component gets notified of property changes via the `onProps` function.

What you'll be looking for in your child component would be:
```
window.xchild.onProps(function (props) {
  console.log(props.counter);
});
```

cmackenz:
Thanks @freshfunkee, that's what I was looking for!

PR #479, opened 2026-03-31
feat: add bfcacheEnabled component option and gate parent guards in zoid

Vishala230531:
Add bfcacheEnabled as a first-class zoid component option so checkout-components can pass it through from the ELMO experiment. When enabled, parent-side watchForUnload skips destroy on bfcache entry, adds a pageshow listener for restore detection, and passes bfcacheAware to belter's watchElementForClose. When disabled (default), behavior is identical to current production.

[https://paypal.atlassian.net/browse/DTPPCPSDK-5699](url)

Screenshot 2026-03-25 at 9 38 50 AM

### DIFF
--- a/src/child/child.js
+++ b/src/child/child.js
@@ -115,6 +115,7 @@ export function childComponent<P, X, C, ExtType>(
     exports: parentExports,
     context,
     props: initialProps,
+    enableBfcache,
   } = payload;
 
   if (!versionCompatabilityCheck(version, __ZOID__.__VERSION__)) {
@@ -228,7 +229,7 @@ export function childComponent<P, X, C, ExtType>(
 
     if ("onpagehide" in window) {
       window.addEventListener("pagehide", (event) => {
-        if (event.persisted) {
+        if (event.persisted && enableBfcache) {
           return;
         }
         checkClose.fireAndForget();
--- a/src/child/child.js
+++ b/src/child/child.js
@@ -227,7 +227,10 @@ export function childComponent<P, X, C, ExtType>(
     });
 
     if ("onpagehide" in window) {
-      window.addEventListener("pagehide", () => {
+      window.addEventListener("pagehide", (event) => {
+        if (event.persisted) {
+          return;
+        }
         checkClose.fireAndForget();
       });
     } else {
--- a/src/component/component.js
+++ b/src/component/component.js
@@ -144,6 +144,8 @@ export type ComponentOptionsType<P, X, C, ExtType> = {|
   children?: () => C,
 
   exports?: ExportsDefinition<X>,
+
+  bfcacheEnabled?: boolean,
 |};
 
 export type AttributesType = {|
@@ -193,6 +195,8 @@ export type NormalizedComponentOptionsType<P, X, C, ExtType> = {|
   children: () => C,
 
   exports: ExportsMapperDefinition<X>,
+
+  bfcacheEnabled: boolean,
 |};
 
 export type ZoidComponentInstance<P, X = void, C = void, ExtType = void> = {|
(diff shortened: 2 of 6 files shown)


src/component/component.js
```diff
@@ -144,6 +144,8 @@ export type ComponentOptionsType<P, X, C, ExtType> = {|
   children?: () => C,
 
   exports?: ExportsDefinition<X>,
+
+  bfcacheEnabled?: boolean,
```
nikrom17: Let's rename `bfcacheEnabled` to `enableBfcache` across the component option, types, and `normalizeOptions`. This follows the `enable*` convention for opt-in flags and reads better at the consumer call site:

```js
// PCC component.jsx
enableBfcache: Boolean(getButtonExperiments()?.isBfcacheEnabled),
```

The rename applies to `ComponentOptionsType`, `NormalizedComponentOptionsType`, `normalizeOptions`, and the destructure in `parent.js`.

src/child/child.js
```diff
@@ -227,7 +227,10 @@ export function childComponent<P, X, C, ExtType>(
     });
 
     if ("onpagehide" in window) {
-      window.addEventListener("pagehide", () => {
+      window.addEventListener("pagehide", (event) => {
+        if (event.persisted) {
```
nikrom17: This skips `checkClose` on `event.persisted` for ALL zoid components, not just those with `bfcacheEnabled: true`. The parent side is properly gated behind `bfcacheEnabled`, but the child side is not.

For components that don't opt into bfcache, the parent will call `destroy()` on bfcache entry (since `bfcacheEnabled` is false), but the child skips `checkClose`, so parent and child teardown signals are inconsistent.

In practice this is likely harmless since the parent handles destroy regardless, but it breaks the pattern. The child doesn't currently have access to `bfcacheEnabled`, so gating would require passing it down via props or the init handshake. Worth noting as a follow-up if we see issues.

src/parent/parent.js
```diff
@@ -1231,19 +1229,23 @@ export function parentComponent<P, X, C, ExtType>({
...
+              return ZalgoPromise.delay(1).then(() => {
+                if (isElementClosed(innerContainer)) {
+                  close(removeError);
+                } else {
+                  clean.all(removeError);
+                  return rerender().then(resolveInitPromise, rejectInitPromise);
+                }
+              });
+            },
+            { bfcacheAware: bfcacheEnabled }
```
nikrom17: Let's rename `bfcacheAware` to `isBfcacheEnabled` to match the ELMO experiment flag name. This keeps naming consistent across the full chain: ELMO -> PCC -> zoid -> belter.

```js
{ isBfcacheEnabled: enableBfcache }
```

There's a corresponding comment on the belter PR (#130) for the receiving side.

nikrom17 (changes requested):

nikrom17:
Pushed a fix commit addressing the outstanding item: \`@krakenjs/belter\` was pinned to the throwaway \`2.10.0-alpha.4\` from the earlier test cycle. Belter's real \`bfcacheAware\`/\`isBfcacheEnabled\` fix has since been merged and published as \`2.10.1\`, so this now points there via \`^2.10.1\`. Also dropped the stray \`10.6.0-alpha.1\` version bump and alpha-cycle CHANGELOG entries that had leaked in from the alpha package testing scripts.

Separately, #478 has been re-applied in #485 (it was reverted for the zoid-wide global-key/version-skew issue fixed in #484 (10.5.3), unrelated to this PR's content) — this PR should merge after #485 so both land in the same release.

Ready for another review pass.

nikrom17 (approved):

src/child/child.js
```diff
@@ -227,7 +228,10 @@ export function childComponent<P, X, C, ExtType>(
     });
 
     if ("onpagehide" in window) {
-      window.addEventListener("pagehide", () => {
+      window.addEventListener("pagehide", (event) => {
+        if (event.persisted && enableBfcache) {
+          return;
```
ravishekhar, line 233: nit, we should add a code comment to explain the early exit.

ravishekhar (approved):
left one non blocking comment, looks good otherwise.

PR #483, opened 2026-06-10
[Feature] Re-render PayPal Button on `onCancel` and `onError`

jtanya17:
### Description

Updates Zoid rerender behavior so a component can be re-rendered into its original container with decorated props.

**Problem:** In app switch resume cancel flows, PayPal Checkout needs to render a fresh button after the merchant `onCancel` callback while preserving resume-specific props, such as reusing the returned `orderID` for retry. Zoid owned the original render target, but `rerender()` could only clone with the original props.

**Solution:** Extend Zoid `rerender()` to accept an optional prop decorator. The rerender path still uses Zoid’s existing target/container/context handling, while callers can override selected props for the new instance.

**Impact:** PayPal Checkout can restore the button after app switch cancel flows without hardcoding merchant container selectors or moving render-target logic into PPC.

### Why are we making these changes?

  1. DTRIMAC-2082
  2. DTPPCPSDK-4375

### Reproduction Steps

Same reproduction flow as the paired paypal-checkout-components PR: https://github.com/paypal/paypal-checkout-components/pull/2592.

### Dependent Changes

To be tested with the paired paypal-checkout-components PR: https://github.com/paypal/paypal-checkout-components/pull/2592 and SCNW PR: #1183.

### DIFF
--- a/src/component/component.js
+++ b/src/component/component.js
@@ -32,6 +32,7 @@ import {
   type ParentComponent,
   type RenderOptionsType,
   type ParentHelpers,
+  type RerenderOptions,
   parentComponent,
 } from "../parent/parent";
 import {
@@ -201,7 +202,9 @@ export type ZoidComponentInstance<P, X = void, C = void, ExtType = void> = {|
   ...X,
   ...C,
   isEligible: () => boolean,
-  clone: () => ZoidComponentInstance<P, X, C, ExtType>,
+  clone: (
+    options?: RerenderOptions<P>
+  ) => ZoidComponentInstance<P, X, C, ExtType>,
   render: (
     container?: ContainerReferenceType,
     context?: $Values<typeof CONTEXT>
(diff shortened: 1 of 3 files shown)


tmanderson (approved):

src/component/component.js
```diff
@@ -384,6 +387,175 @@ export function component<P, X, C, ExtType>(
   const global = getGlobal(window);
   const driverCache = {};
   const instances = [];
+  let latestRenderedRerender: ?(
```
nikrom17, line 390: `latestRenderedRerender` and the sessionStorage key on line 393 are scoped per-tag, shared by every instance from the same `zoid.create()` call (same scope as `instances` on line 389), not per-instance.

Scenario with two instances of the same tag:
1. Instance A renders into container A - `writeLatestRender` stores container A's selector under `__zoid_latest_render__<tag>`.
2. Instance B renders into container B - overwrites the same key with container B's selector.
3. Page reloads. A new instance representing A calls `.rerender()` directly, without `.render()` first.
4. It has no in-memory `currentRerender`, so it falls to `getFallbackRerender()`, which reads the one sessionStorage entry - now pointing at B.

Undesired behavior: A's rerender lands in container B instead of A.

Possible fix: key the sessionStorage entry (and `latestRenderedRerender`) per-instance rather than per-tag, e.g. a uid generated at first render.

src/parent/parent.js
```diff
@@ -317,6 +325,9 @@ export function parentComponent<P, X, C, ExtType>({
   let childComponent: ?ChildExportsType<P>;
   let currentChildDomain: ?string;
   let currentContainer: HTMLElement | void;
+  let currentRerender: ?Rerender<P> = null;
```
nikrom17, line 328: `currentRerender`/`lastRerender` are set in `render()` (line 1539) and never cleared, including in `destroy()`.

Scenario:
1. Component renders - `currentRerender`/`lastRerender` get set, `hasBeenRendered` becomes true.
2. Caller calls `.close()`, which internally calls `destroy()`. Nothing resets `currentRerender`/`lastRerender`.
3. Caller calls `.rerender()` again on the same instance.
4. `getHelpers().rerender` finds `currentRerender` still set, so it calls it directly - the guard below only throws when `currentRerender`, `lastRerender`, and the fallback are all falsy, which is never true once the component has rendered once:
```js
if (!hasBeenRendered) { throw new Error("...must be rendered first."); }
throw new Error("Rerender callback lost after render - component may be destroyed or re-initialized.");
```

Undesired behavior: rerender on a destroyed component silently succeeds and re-renders it, instead of throwing the "component may be destroyed or re-initialized" error this is meant to produce.

Possible fix: reset `currentRerender = null; lastRerender = null;` inside `destroy()`.

src/parent/parent.js
```diff
@@ -1507,8 +1534,11 @@ export function parentComponent<P, X, C, ExtType>({
     container,
     context,
     rerender,
-  }: RenderOptions): ZalgoPromise<void> => {
+  }: RenderOptions<P>): ZalgoPromise<void> => {
     return ZalgoPromise.try(() => {
+      currentRerender = rerender;
```
nikrom17, line 1539:
```js
currentRerender = rerender; lastRerender = rerender; hasBeenRendered = true;
```

These run at the very top of `render()`'s promise chain, before the rest of initialization executes. If `render()` ultimately fails partway through, the component still looks "renderable" to `getHelpers().rerender()`. Should these move to after render actually succeeds?

src/component/component.js
```diff
@@ -384,6 +387,175 @@ export function component<P, X, C, ExtType>(
...
+      }
+
+      const elementName = container.getAttribute("name");
+      if (elementName) {
+        persistedContainers.push(`[name="${escapeAttr(elementName)}"]`);
+      }
+
+      const className = container.className;
+      if (typeof className === "string" && className.trim()) {
+        const firstClass = className.trim().split(/\s+/)[0];
```
nikrom17, line 519: Only the first CSS class gets kept as a candidate selector here, and `readLatestRender` falls back to `candidates[0]` (line 485) even when nothing in the current DOM matches. In apps using utility-first CSS (Tailwind etc.) the first class is rarely the meaningful one, and frameworks that remount DOM nodes on navigation will lose the `data-zoid-latest-container-*` marker attribute too. Worst case this matches an unrelated element that happens to share a class/id, or targets a selector that means nothing anymore. Haven't verified this against real merchant container markup, just flagging the pattern as fragile.

src/component/component.js
```diff
@@ -518,9 +690,31 @@ export function component<P, X, C, ExtType>(
...
     };
 
+    const clone = ({ decorate = identity }: RerenderOptions<P> = {}) => {
+      return init(decorate(props));
+    };
+
     const parent = parentComponent({
       uid,
       options,
+      getFallbackRerender: () => {
```
nikrom17, line 700: `getFallbackRerender` is wired in unconditionally here, so every component writes to sessionStorage on every render (`writeLatestRender`) whether or not the consumer ever calls `.rerender()`.

Given the container-matching in `readLatestRender`/`waitForSelector` is inherently heuristic (see comment above on selector matching), would it make sense to gate this behind a `zoid.create()` option, e.g. `enableRerenderFallback`? That way components that don't need cross-navigation rerender skip the sessionStorage write and the fragile-match failure mode entirely.

test/tests/rerender.js
```diff
@@ -116,6 +116,64 @@ describe("zoid rerender cases", () => {
     });
   });
 
+  it("should re-render a component with decorated props", () => {
```
nikrom17, line 119: This test covers the simple in-memory case (decorate props, rerender on the same page), but there's nothing here for `getFallbackRerender`, `writeLatestRender`/`readLatestRender`, or `waitForSelector`. That's the entire persistence-across-navigation mechanism, and it's the highest-risk, most novel part of this change. Worth adding a test that simulates a fresh page load (new component instance, never rendered, `rerender()` called directly) and asserts it finds the right container via sessionStorage, plus a test with two instances of the same tag to catch the cross-instance issue flagged above.

